PR #15, opened 2025-05-07
feat(assembly): add packaging functionality

venetrius:
related to: camunda/camunda-bpm-platform/issues/4988


assembly/resources/application.yml
```diff
@@ -0,0 +1,13 @@
+migrator:
+  rdbms-exporter:
+    auto-ddl: true
+  source:
+    jdbc-url: jdbc:h2:~/Downloads/c8_rdbms-source.db;TRACE_LEVEL_FILE=0;DB_CLOSE_ON_EXIT=FALSE
```
venetrius, line 5: I was wondering removing this hard coded config. It would make no sense to add in memory h2 here. If someone runs the `examples/generate-runtime` this is where the files are created.

distro/src/main/java/io/camunda/migrator/app/MigratorApp.java
```diff
@@ -0,0 +1,54 @@
...
+@SpringBootApplication
+public class MigratorApp {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(MigratorApp.class);
+
+  public static void main(String[] args) {
+    ConfigurableApplicationContext context = SpringApplication.run(MigratorApp.class, args);
+    ApplicationArguments appArgs = new DefaultApplicationArguments(args);
+    try {
+      if (appArgs.containsOption("runtime")) {
```
tasso94, line 29: /nitpick We could move this and "history" constants into `protected static final` fields.

tasso94 (changes requested):
For the sake of demonstration it would be fun to add a custom Data Migrator Spring Boot banner instead of using the default one.

image

assembly/assembly.xml
```diff
@@ -0,0 +1,77 @@
...
+      <includes>
+        <include>com.h2database:h2</include>
+      </includes>
+    </dependencySet>
+  </dependencySets>
+
+   <files>
+  <file>
+    <source>${project.build.directory}/dependency/c7-data-migrator-distro-${project.version}.jar</source>
+    <outputDirectory>lib</outputDirectory>
```
tasso94, line 36: /suggest Can we name this folder `internal` to make it transparent that the user shouldn't mess with what's inside here? Libs are included in the Uber jar anyway, so the name `lib` is even a bit misleading, and users might be confused by the purpose of `lib` and `userlib' folders.

assembly/assembly.xml
```diff
@@ -0,0 +1,77 @@
...
+      <fileMode>0755</fileMode>
+    </file>
+    <file>
+      <source>resources/README.txt</source>
+      <filtered>true</filtered>
+      <outputDirectory></outputDirectory>
+      <fileMode>0755</fileMode>
+    </file>
+    <file>
+      <source>../CAMUNDA-LICENSE-1.0.txt</source>
```
tasso94, line 63: /suggest Can we just name this file `LICENSE.txt` to make it unambiguously what this is about?

assembly/assembly.xml
```diff
@@ -0,0 +1,77 @@
...
+      <outputDirectory>lib</outputDirectory>
+      <useProjectArtifact>false</useProjectArtifact>
+      <useTransitiveDependencies>false</useTransitiveDependencies>
+      <excludes>
+        <exclude>io.camunda:c7-data-migrator-distro</exclude>
+        <exclude>com.h2database:h2</exclude>
+      </excludes> 
+    </dependencySet>
+    <dependencySet>
+      <outputDirectory>userlib/</outputDirectory>
```
tasso94, line 26: /suggest Can we move the `userlib` folder into the `configuration` folder? Like this, the user can find everything they can control in one folder. This would also be consistent with Camunda Run, even though that's not a hard requirement. But I think it makes sense; otherwise, I would challenge why we would need a single file, `application.yml` in a dedicated `configuration` folder. This indirection wouldn't create much value. However, when including all user-owned, this folder makes much sense.

assembly/resources/README.txt
```diff
@@ -0,0 +1,48 @@
...
+2. Execute the `start` script:  
+   - `start.bat` for Windows  
+   - `start.sh` for Linux/macOS
+
+Options
+
+- `--runtime` – Migrate runtime data only  
+- `--history` – Migrate history data only 
+
+==================
```
tasso94, line 27: /suggest Would be cool to have a more general get started guide in the `README.md` in the root of this project.

assembly/resources/application.yml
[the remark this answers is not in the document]
venetrius: If I update the url to:
```
jdbc-url: jdbc:h2:./c8_rdbms-source.db;TRACE_LEVEL_FILE=0;DB_CLOSE_ON_EXIT=FALSE
```
It works. The downside is that now I to update [`examples/generate-runtime/src/main/resources/application.yml`](https://github.com/camunda/c7-data-migrator/blob/main/examples/generate-runtime/src/main/resources/application.yml) before generating data to point to where the script is called from.

I think it is important for the user to able to use the app seamlessly. If you agree I will push the related changes.

assembly/resources/start.sh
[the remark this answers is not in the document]
venetrius: I would prefer this to be tackled in a follow up task. If it ok, I will create one.

assembly/resources/README.txt
venetrius: Yana already updated the README addressing some of this points. I will create a new PR to update based on the rest. I am creating a new PR as the README changes are already in main.

assembly/assembly.xml
venetrius: Updated. Changes: https://github.com/camunda/c7-data-migrator/pull/15/commits/64a63a9d261c92cb1a51c0985db2611d7fcdf057

assembly/assembly.xml
venetrius: Good point. Updated https://github.com/camunda/c7-data-migrator/pull/15/commits/515744651a1c8327ee29f29c50649dbad9423201

assembly/assembly.xml
venetrius: Good point, updated. Changes https://github.com/camunda/c7-data-migrator/pull/15/commits/98a60eaec48eb84dd14fdc60f12230f017dc2261

distro/src/main/java/io/camunda/migrator/app/MigratorApp.java
venetrius: Good point, updated. https://github.com/camunda/c7-data-migrator/pull/15/commits/4a6e4cc634620313aa53eea8b2c76f10433f154a

assembly/resources/README.txt
```diff
@@ -0,0 +1,48 @@
+This is a distribution of
+
+       Camunda 7 Data Migrator v${project.version}
+
+visit
+       http://docs.camunda.org/
```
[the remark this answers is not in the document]
venetrius: Good point. Updated https://github.com/camunda/c7-data-migrator/pull/15/commits/ae58cb42fc8363d54b3b97480da45d2a7c6f5533

assembly/pom.xml
```diff
@@ -0,0 +1,100 @@
...
+        </configuration>
+        <executions>
+          <execution>
+            <phase>verify</phase>
+            <goals>
+              <goal>check</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
```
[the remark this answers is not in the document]
venetrius: The license check fails otherwise.
The root pom already has this defined. I was not able to simply fit is quickly. I am going to open a follow up issue.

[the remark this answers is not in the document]
tasso94: I can see that in the root pom its configured in the `pluginManagement` section instead of the `plugins` section. `pluginManagement` needs to be defined again in each child module, whereas `plugins` section is inherited automatically by all child modules.

venetrius:
Created follow up tasks: 
- [In Data migrator - add a custom Data Migrator Spring Boot banner #5141
](https://github.com/camunda/camunda-bpm-platform/issues/5141)
- [In Data migrator - remove unnecessary libraries from bundeled app](https://github.com/camunda/camunda-bpm-platform/issues/5142)
- [In Data migrator - simplify license check](https://github.com/camunda/camunda-bpm-platform/issues/5143)
[- In Data migrator - assembly enable scripts to run runtime / history in user defined order 
](https://github.com/camunda/camunda-bpm-platform/issues/5144)

Todo from my side: update root Readme.

> When I'm not in the directory where the start.sh script lives, I get the following error. Would it be possible to run the start script even when not inside the directory?
> % ./c7-data-migrator-assembly-0.0.1-SNAPSHOT/start.sh 
> Starting application without migration flag.
> Error: Unable to access jarfile ./lib/c7-data-migrator.jar

Good call. Updated to be able to run script from other dirs as well

tasso94 (approved):
Changes and follow up tickets look good to me 👍

PR #24, opened 2025-05-13
task(assembly): setup log output

venetrius:
related to: https://github.com/camunda/camunda-bpm-platform/issues/4987

### DIFF
--- a/core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
+++ b/core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
@@ -106,10 +106,15 @@ public class RuntimeMigrator {
       LOGGER.info("Migrating process instance with legacyId [{}]", legacyProcessInstanceId);
       Map<String, Object> globalVariables = new HashMap<>();
 
-      runtimeService.createVariableInstanceQuery()
-          .activityInstanceIdIn(legacyProcessInstanceId)
-          .list()
-          .forEach(variable -> globalVariables.put(variable.getName(), variable.getValue())); // Collectors#toMap cannot handle null values and throws NPE.
+      try {
+        runtimeService.createVariableInstanceQuery()
+            .activityInstanceIdIn(legacyProcessInstanceId)
+            .list()
+            .forEach(variable -> globalVariables.put(variable.getName(), variable.getValue())); // Collectors#toMap cannot handle null values and throws NPE.
+      } catch (Exception e) {
+        LOGGER.error("Error while querying legacy global variables {}", e.getMessage());
+        throw new RuntimeException(e);
+      }
 
       globalVariables.put("legacyId", legacyProcessInstanceId);
       LOGGER.debug("Global variables for process  instance [{}] set to [{}]", legacyProcessInstanceId, globalVariables);
@@ -137,14 +142,19 @@ public class RuntimeMigrator {
       LOGGER.debug("Activating jobs");
       List<ActivatedJob> migratorJobs = null;
       do {
-        migratorJobs = camundaClient.newActivateJobsCommand()
-            .jobType("migrator")
-            // TODO: review #maxJobsToActivate and #timeout
-            .maxJobsToActivate(Integer.MAX_VALUE)
-            .timeout(Duration.ofMinutes(1))
-            .send()
-            .join()
-            .getJobs();
+        try {
+          migratorJobs = camundaClient.newActivateJobsCommand()
+              .jobType("migrator")
+              // TODO: review #maxJobsToActivate and #timeout
+              .maxJobsToActivate(Integer.MAX_VALUE)
+              .timeout(Duration.ofMinutes(1))
+              .send()
+              .join()
+              .getJobs();
+        } catch (Exception e) {
+          LOGGER.error("Error while activating jobs {}", e.getMessage());
+          throw new RuntimeException(e);
+        }
 
         LOGGER.debug("Preparing to migrate {} jobs", migratorJobs.size());
         migratorJobs.forEach(activatedJob -> {
@@ -183,8 +193,14 @@ public class RuntimeMigrator {
                 modifyInstructions = modifyProcessInstance.activateElement(activityId)
                     .withVariables(localVariables, activityId);
               }
-              modifyInstructions.send().join();
-              // no need to complete the job since the modification canceled the migrator job in the start event
+
+              try {
+                modifyInstructions.send().join();
+                // no need to complete the job since the modification canceled the migrator job in the start event
+              } catch (Exception e) {
+                LOGGER.error("Error while modifying jobs {}", e.getMessage());
+                throw new RuntimeException(e);
+              }
             });
       } while (!migratorJobs.isEmpty());
       LOGGER.info("Migration completed for process instance with migrated key [{}] and legacyId [{}]",


tasso94 (commented):
/q Do we need logging for potential errors?

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
```diff
@@ -74,14 +79,18 @@ public void migrate() {
       if (deployResourceCommandStep2 != null) {
         deployResourceCommandStep2.send().join();
       }
+      LOGGER.debug("Completed auto deployment.");
     }
 
     String latestLegacyId = idKeyMapper.findLatestIdByType("runtimeProcessInstance");
+    LOGGER.debug("Latest legacy id is {}", latestLegacyId);
```
joaquinfelici, line 86: 🙃 Maybe we could be a bit more precise here:
```suggestion
    LOGGER.debug("Legacy ID of latest migrated process instance: [{}]", latestLegacyId);
```

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
```diff
@@ -74,14 +79,18 @@ public void migrate() {
...
     ProcessInstanceQuery processInstanceQuery = ((ProcessInstanceQueryImpl) runtimeService.createProcessInstanceQuery())
         .idAfter(latestLegacyId)
         .rootProcessInstances();
 
     long maxLegacyProcessInstanceCount = processInstanceQuery.count();
+    LOGGER.debug("Number of legacy process instances to migrate is {}", maxLegacyProcessInstanceCount);
```
joaquinfelici, line 92: 🙃 Just an idea.
```suggestion
    LOGGER.debug("{} process instances will be migrated", maxLegacyProcessInstanceCount);
```
or:
```suggestion
    LOGGER.debug("{} process instances are scheduled for migration", maxLegacyProcessInstanceCount);
```

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
```diff
@@ -95,8 +104,10 @@ public void migrate() {
 
     // TODO: paginate this
     idKeyMapper.findProcessInstanceIds().forEach(legacyProcessInstanceId -> {
+      LOGGER.info("Migrating Process instance with legacyId {}", legacyProcessInstanceId);
```
joaquinfelici, line 107:
```suggestion
      LOGGER.info("Migrating process instance with legacyId [{}]", legacyProcessInstanceId);
```

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
```diff
@@ -109,6 +120,7 @@ public void migrate() {
           .singleResult()
           .getProcessDefinitionKey();
 
+      LOGGER.debug("Creating new process instance");
```
joaquinfelici, line 123: I think we could try to be consistent with ending sentences with a period or not, I read somewhere that it's best practice **not** to include a period, but I don't have a strong opinion, just that we're consistent.

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
```diff
@@ -95,8 +104,10 @@ public void migrate() {
 
     // TODO: paginate this
     idKeyMapper.findProcessInstanceIds().forEach(legacyProcessInstanceId -> {
+      LOGGER.info("Migrating Process instance with legacyId {}", legacyProcessInstanceId);
       Map<String, Object> globalVariables = new HashMap<>();
 
+      LOGGER.debug("Loading legacy global process instance variables.");
```
joaquinfelici, line 110: I think it makes more sense to log the result of the globalVariables map below. Something like:

```Global variables for process  instance [{}] set to [{}]```.

File: core/src/main/java/io/camunda/migrator/RuntimeMigrator.java [hunk 6]
```diff
@@ -117,11 +129,14 @@ public void migrate() {
           .join()
           .getProcessInstanceKey();
 
+      LOGGER.debug("Created new process instance with key {} for legacyProcessInstanceId {}", processInstanceKey,
```
joaquinfelici, line 132: ❓ I feel like this deserves INFO level, what do you think?

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
```diff
@@ -132,7 +147,7 @@ public void migrate() {
             .send()
             .join()
             .getJobs();
-            migratorJobs.forEach(activatedJob -> {
```
joaquinfelici, line 135: ❓ Maybe we could log how many migrator jobs were found.

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
```diff
@@ -172,6 +187,8 @@ public void migrate() {
               // no need to complete the job since the modification canceled the migrator job in the start event
             });
       } while (!migratorJobs.isEmpty());
+      LOGGER.info("Migration completed for Process instance with legacyId {}, the key of the migrated process is {}",
```
joaquinfelici, line 190: 🙃 Just a suggestion of phrasing here to make it less verbose.
```suggestion
      LOGGER.info("Migration completed for process instance with migrated key [{}] and legacyId [{}]",
```

joaquinfelici (changes requested):
Looks good overall! 👍 
Added some suggestions and question.

joaquinfelici:
I also agree with @tasso94 above that logging error situations could be useful. I feel like this may require some code changes though.

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
```diff
@@ -74,14 +79,18 @@ public void migrate() {
       if (deployResourceCommandStep2 != null) {
```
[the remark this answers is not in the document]
venetrius: I think it is not necessary. We want to remove this feature, it is only here to make the development process easier.

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
venetrius: Good point! Removed periods

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
venetrius: Updated to info

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
venetrius: Updated

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
venetrius: Updated. 
Do you think it can be an issue if there are a lot of variables? The log file will be huge but it is only for debug mode, should be okay.
What about security? Is it ok to log out all variables?

joaquinfelici: Interesting point, maybe we can log a list of variable keys (by using `globalVariables.keySet()`) but not the values. And yes I think that if it's a debug log it's useful to see what has been migrated, even if it's a big file.

venetrius:
> I also agree with @tasso94 above that logging error situations could be useful. I feel like this may require some code changes though.

Added error logs, where I felt they are useful.

venetrius:
Will close this one and open a new PR as the related code has been refactored recently

PR #29, opened 2025-05-15
feat(core): make C8 client configurable

tasso94:
* Makes the C8 client configurable via `application.yml` file.
* Introduces a new flag `migrator.c7.auto-ddl` to control if C7 schema should be created automatically.
* Configures source as primary data source.
* Configures source and target data sources to the same in-memory database for easier getting started/test experience.
* Refactors `qa/` test setup to run tests on in-memory db.
* Adds tests.
* Fix a compilation issue due to an update of the C8 RDBMS module.

related to camunda/camunda-bpm-platform#5000


core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
```diff
@@ -254,8 +253,8 @@ public static List<Path> findFilesInDirectory(File directory, String extension)
     return result;
   }
 
-  public void setCamundaClient(CamundaClient camundaClient) {
-    this.camundaClient = camundaClient;
+  public CamundaClient getCamundaClient() {
```
joaquinfelici, line 256: ❓ Do we really need this or can the client simply be injected in the tests?

core/pom.xml
```diff
@@ -96,7 +96,35 @@
     </dependency>
 
     <!-- Test Dependencies -->
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+    </dependency>
```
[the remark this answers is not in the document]
tasso94: Good catch! 👍

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
tasso94: Probably not. 👍

qa/src/test/java/io/camunda/migrator/qa/RuntimeMigrationAbstractTest.java
```diff
@@ -11,28 +11,31 @@
...
 
   @Autowired
   protected RepositoryService repositoryService;
 
+  protected CamundaClient camundaClient;
+
   @PostConstruct
   public void init() {
-    runtimeMigrator.setCamundaClient(camundaClient);
+    camundaClient = runtimeMigrator.getCamundaClient();
```
joaquinfelici, line 38: 🔧 We also need to undo these changes so that we inject the client

joaquinfelici (approved):
Approved 👍 
Just one comment below.

PR #31, opened 2025-05-21
feat(distro): allow C8 resource auto-deployment

tasso94:
related to camunda/camunda-bpm-platform#5156

### DIFF
--- a/core/src/main/java/io/camunda/migrator/AutoDeployer.java
+++ b/core/src/main/java/io/camunda/migrator/AutoDeployer.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.migrator;
+
+import io.camunda.client.CamundaClient;
+import io.camunda.client.api.command.DeployResourceCommandStep1;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Collections;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+import org.springframework.util.StringUtils;
+
+@Component
+public class AutoDeployer {
+
+  @Autowired
+  protected CamundaClient camundaClient;
+
+  @Value("${migrator.deployment-dir:#{null}}")
+  protected String deploymentDir;
+
+  public void deploy() {
+    // Deploy process
+    var deployResource = camundaClient.newDeployResourceCommand();
+
+    Set<Path> models = getDeploymentResources();
+
+    DeployResourceCommandStep1.DeployResourceCommandStep2 deployResourceCommandStep2 = null;
+    for (Path model : models) {
+      deployResourceCommandStep2 = deployResource.addResourceFile(model.toString());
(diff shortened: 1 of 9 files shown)


core/src/main/java/io/camunda/migrator/AutoDeployer.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  public Set<Path> getDeploymentResources() {
+    if (StringUtils.hasText(deploymentDir)) {
+      Path resourceDir = Paths.get(deploymentDir);
+
+      try (Stream<Path> stream = Files.walk(resourceDir)) {
+        return stream.filter(file -> !Files.isDirectory(file)).collect(Collectors.toSet());
+      } catch (IOException e) {
+        e.printStackTrace(); // TODO: Add error logging when available.
```
HeleneW-dot, line 57: Nit:
Is there a ticket number we can associate with this todo?

core/src/main/java/io/camunda/migrator/AutoDeployer.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+import org.springframework.util.StringUtils;
+
+@Component
+public class AutoDeployer {
```
HeleneW-dot, line 26: Nit:
This is only gonna be used for C8 deployments, right? If so I would like the name to be more specific so its clearer this isnt deploying anything else. Maybe something like `C8ProcessAutoDeployer`?

HeleneW-dot (approved):
Lgtm! Tested locally as well

core/src/main/java/io/camunda/migrator/AutoDeployer.java
```diff
@@ -0,0 +1,62 @@
...
+
+    Set<Path> models = getDeploymentResources();
+
+    DeployResourceCommandStep1.DeployResourceCommandStep2 deployResourceCommandStep2 = null;
+    for (Path model : models) {
+      deployResourceCommandStep2 = deployResource.addResourceFile(model.toString());
+    }
+
+    if (deployResourceCommandStep2 != null) {
+      deployResourceCommandStep2.send().join();
```
tasso94, line 46: /q What happens if we restart the migrator without changing the model? Is a new version deployed?

HeleneW-dot: I just tried this out and it appears if I either change a model or add a model, all models are redeployed and get a new version.

core/src/main/java/io/camunda/migrator/AutoDeployer.java
tasso94: Having this in the class name might make understanding what API it calls easier. However, I'm a bit afraid of cluttering the whole migrator code base with C7 and C8 prefixes and I'm unsure if this really provides much value. We might extend this class in the future with C7 auto-deployment capabilities. Would it mean that we have to rename this class again? What would be the name? C7C8AutoDeployer. Not sure if that is great TBH. Isn't it relatively straightforward what happens and which APIs are involved by just reading the code?

core/src/main/java/io/camunda/migrator/AutoDeployer.java
tasso94: Made a note to camunda/camunda-bpm-platform#4987

core/src/main/java/io/camunda/migrator/AutoDeployer.java
tasso94: I think that's acceptable for now. Let's wait for user feedback.

core/src/main/java/io/camunda/migrator/AutoDeployer.java
HeleneW-dot: Yeah thats fair, especially if we plan on extending it.

PR #32, opened 2025-05-22
chore(core): remove multi-instance workaround

tasso94:
related to camunda/camunda-bpm-platform#5176

### DIFF
--- a/core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
+++ b/core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
@@ -27,8 +27,6 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
-import java.util.stream.Collectors;
 
 import static io.camunda.migrator.HistoryMigrator.BATCH_SIZE;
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
```diff
@@ -144,22 +139,6 @@ public void migrate() {
...
-  /**
-   * This removes multi-instances from the active activity instance map and only keeps the #multiInstanceBody activity.
-   * Like this, the C8 process model takes care of instantiating X instances.
-   * TODO: how would this work for multi-instance call activities?
-   */
-  protected void removeMultiInstances(Map<String, ActInstance> activityInstanceMap) {
-    Set<String> multiInstanceBodies = activityInstanceMap.values()
-        .stream()
-        .map(ActInstance::id)
-        .filter(activityId -> activityId.endsWith("#multiInstanceBody"))
```
joaquinfelici, line 156: /q that might be useful for https://github.com/camunda/camunda-bpm-platform/issues/4985. Do you know if this enough condition to determine whether the process has multi instance? Just thinking about how to re-use this to create the initial validation check 🤔

joaquinfelici (approved):
Looks good 👍 
Left one (kind of) unrelated question.

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
tasso94: Good question. Ideally, we add a bunch of tests to validate if we can detect it reliably.
In BPMN there is a concept of multi-instance markers. Is such attribute available in the transient Java representation?🤔

tasso94: I have an idea of how to validate any multi-instance in the BPMN.

PR #34, opened 2025-05-26
chore(IT): make test suffix consistent

HeleneW-dot:
See [slack conversation](https://camunda.slack.com/archives/CL5PTD6AJ/p1747991697118729)

Naming all tests `..Test` because they are run automatically by maven build with us using surefire. `..IT` naming would require us to switch to failsafe

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -150,6 +150,11 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>3.5.3</version>
+      </plugin>
     </plugins>
   </build>
 </project>


core/src/test/java/io/camunda/migrator/CamundaClientConfigurationTest.java
```diff
@@ -21,7 +21,7 @@
     "camunda.client.grpc-address=http://helloworld:33333",
     "camunda.client.rest-address=http://helloworld:44444"
 })
-public class CamundaClientConfigurationIT {
+public class CamundaClientConfigurationTest {
```
tasso94, line 24: Looking at the logs, I still see this test not being executed? 🤔 

I think the reason for that is we need to add the surefire plugin to the plugins in the pom.xml of the core module:

```xml
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-surefire-plugin</artifactId>
        <version>3.5.3</version>
      </plugin>
```

Can you take care?

tasso94 (changes requested):

core/src/test/java/io/camunda/migrator/CamundaClientConfigurationTest.java
HeleneW-dot: Added ✔️  Lets see

tasso94 (approved):

PR #42, opened 2025-05-28
feat(runtime): validate runtime state

joaquinfelici:
Related to https://github.com/camunda/camunda-bpm-platform/issues/4999

### DIFF
--- a/qa/pom.xml
+++ b/qa/pom.xml
@@ -14,18 +14,6 @@
 
   <artifactId>c7-data-migrator-qa</artifactId>
 
-  <dependencyManagement>
-    <dependencies>
-      <dependency>
-        <groupId>org.springframework.boot</groupId>
-        <artifactId>spring-boot-dependencies</artifactId>
-        <version>${version.spring-boot}</version>
-        <type>pom</type>
-        <scope>import</scope>
-      </dependency>
-    </dependencies>
-  </dependencyManagement>
-
   <dependencies>
     <dependency>
       <groupId>io.camunda</groupId>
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
```diff
@@ -128,8 +137,29 @@ protected Map<String, Object> generateGlobalVariables(String legacyProcessInstan
...
+    List<ProcessInstance> processInstances = runtimeService.createProcessInstanceQuery().rootProcessInstanceId(legacyProcessInstanceId).list();
+
+    for (ProcessInstance processInstance : processInstances) {
+      ActivityInstance activityInstanceTree = runtimeService.getActivityInstance(processInstance.getId());
+      Map<String, ActInstance> activityInstanceMap = getActiveActivityIdsById(activityInstanceTree, new HashMap<>());
+      BpmnModelInstance bpmnModelInstance = repositoryService.getBpmnModelInstance(processInstance.getProcessDefinitionId());
+
+      for (ActInstance actInstance : activityInstanceMap.values()) {
+        FlowElement element = bpmnModelInstance.getModelElementById(actInstance.activityId());
+        if ((element instanceof Activity activity) && (activity.getLoopCharacteristics() instanceof MultiInstanceLoopCharacteristics)) {
```
joaquinfelici, line 157: Some pointers for this check:
- `instanceof` does null check (so we check if element is null and if loopCharacteristics is null).
- When casting directly to `Activity`, some activityInstances (such as a parallel gateway) cannot be casted, throwing an exception, hence the FlowElement type and the first check.

tasso94 (approved):
Looks good to me 👍 Great job!

PR #43, opened 2025-05-30
chore(distro): disable history migration temporarily

tasso94:
* Let's disable history migration when not providing any flags temporarily since we don't focus on history migration for now, and the getting started experience is degraded when seeing errors coming from the history migrator.
* We will revert this behavior once work on the history migrator starts.

related to camunda/camunda-bpm-platform#5189


joaquinfelici (approved):
👍

PR #44, opened 2025-06-02
chore(qa): add calling activity element tests

venetrius:
related to: camunda/camunda-bpm-platform/issues/5167


qa/src/test/java/io/camunda/migrator/qa/SubprocessMigrationTest.java
```diff
@@ -0,0 +1,46 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.provider.Arguments;
+
+public class SubprocessMigrationTest extends RuntimeMigrationAbstractTest{
+
+  @Test
+  public void migrateCallActivityAndSubprocess() {
+    // given
+    deployProcessInC7AndC8("calledProcessInstance.bpmn");
+    deployProcessInC7AndC8("callActivity.bpmn");
```
HeleneW-dot, line 28: Nit
the other bpmns have names line `...Process.bpmn`, could rename these cor consistency. Eg `
```suggestion
    deployProcessInC7AndC8("callActivitySubprocess.bpmn");
    deployProcessInC7AndC8("callActivityProcess.bpmn");
```

qa/src/test/java/io/camunda/migrator/qa/SubprocessMigrationTest.java
```diff
@@ -0,0 +1,46 @@
...
+        .createProcessInstanceQuery()
+        .superProcessInstanceId(instance.getProcessInstanceId())
+        .singleResult();
+    // when
+    runtimeMigrator.migrate();
+
+    // then
+    assertThat(byProcessId("callingProcessId")).isActive()
+        .hasActiveElements(byId("callActivityId"))
+        .hasVariable("legacyId", instance.getProcessInstanceId());
```
HeleneW-dot, line 40:
```suggestion
        .hasVariable(LEGACY_ID_VAR_NAME, instance.getProcessInstanceId());
```

README.md
```diff
@@ -21,6 +21,14 @@
   - depending on your implementation, you may need to add [a correlation variable](https://docs.camunda.io/docs/components/modeler/bpmn/message-events/#messages) to the instance pre migration
 - Boundary events:
   - currently not supported
+- Call Activity
+   - To migrate a subprocess that is started from a call activity, the migrator must set the `legacyId` variable for the subprocess. This requires propagating the parent variables. This can be achieved by updating the C8 call activity in one of the following ways:
```
HeleneW-dot, line 25: Question: are both instances truly migrated or are we just migrating the parent instance and then C8 automatically creates the child instance and it moves naturally through the subprocess?

qa/src/test/java/io/camunda/migrator/qa/SubprocessMigrationTest.java
```diff
@@ -0,0 +1,46 @@
...
+    assertThat(byProcessId("calledProcessInstanceId")).isActive()
```
HeleneW-dot, line 41: Question: there is a `rootProcessInstanceId` field on `ProcessInstance`, would that field have the parent Id for child process instance here? If so, can we also add an assert on that

HeleneW-dot (approved):
Lgtm, just have some questions for my understanding of how this migration works. Do we not support the event subprocess or will there be another PR? If so lets also add that to the readme

qa/src/test/java/io/camunda/migrator/qa/SubprocessMigrationTest.java
HeleneW-dot: Ah interesting. So this will not be set during migration then? Is this at all relevant for execution purposes?

README.md
venetrius: We only create the root process, the subprocess is created by C8.
My understanding is:
- We create the parent issue via C8 API
- We activate the parent issue's call activity -> this makes C8 create the subprocess
- We  activate the subprocess

qa/src/test/java/io/camunda/migrator/qa/SubprocessMigrationTest.java
venetrius: Given that the subprocess is created by C8 when the parent process call activity is activated we don't need to set it.

venetrius:
The CI passed for this branch, not sure why it is not shown: https://github.com/camunda/c7-data-migrator/actions/runs/15488900207/job/43609740454

qa/src/test/java/io/camunda/migrator/qa/SubprocessMigrationTest.java
venetrius: Updated file names to `calledActivitySubprocess` & `callActivityProcess`

venetrius:
I will create a new PR for testing Event subprocess

HeleneW-dot (approved):
Approving since the latest CI run passed: https://github.com/camunda/c7-data-migrator/actions/runs/15488900207/job/43609740454

@venetrius do you also want to create a subticket for that followo up PR so we can track it properly?

PR #45, opened 2025-06-03
refactor(core): wrap known exceptions

tasso94:
* Wraps known exceptions in a `RuntimeMigratorException`.
* Creates a builder to simplify pagination.

related to camunda/camunda-bpm-platform#5193


core/src/test/java/io/camunda/migrator/RuntimeMigratorTest.java
tasso94, line 1: Unfortunately, these mocked tests break easily after the implementation is changed slightly. I attempted to fix them in a time box but ran out of time. 🙁

core/src/main/java/io/camunda/migrator/ExceptionUtils.java
```diff
@@ -0,0 +1,56 @@
...
+  /**
+   * Wraps {@link ClientException}, {@link ProcessEngineException}, and {@link PersistenceException} into {@link RuntimeMigratorException}.
+   */
+  public static <T> T callApi(Supplier<T> supplier, String message) {
+    try {
+      return supplier.get();
+    } catch (ClientException | ProcessEngineException | PersistenceException e) {
+      var exception = new RuntimeMigratorException(message, e);
+      LOGGER.error(message, exception);
+      throw exception;
```
tasso94, line 34: Wrapping exception first and next logging the wrapped exception ensures the log output looks trustful for the user as something expected, coming from the migrator:

```
2025-06-03T08:33:57.113+02:00 ERROR 69273 --- [           main] io.camunda.migrator.ExceptionUtils       : Error occurred: shutting down Data Migrator gracefully.

io.camunda.migrator.RuntimeMigratorException: Error occurred: shutting down Data Migrator gracefully.
	at io.camunda.migrator.ExceptionUtils.callApi(ExceptionUtils.java:32) ~[c7-data-migrator-core-0.1.0-SNAPSHOT.jar!/:0.1.0-SNAPSHOT]
	at io.camunda.migrator.ExceptionUtils.callApi(ExceptionUtils.java:22) ~[c7-data-migrator-core-0.1.0-SNAPSHOT.jar!/:0.1.0-SNAPSHOT]
	at io.camunda.migrator.RuntimeMigrator.activateMigratorJobs(RuntimeMigrator.java:201) ~[c7-data-migrator-core-0.1.0-SNAPSHOT.jar!/:0.1.0-SNAPSHOT]
	at io.camunda.migrator.RuntimeMigrator.migrate(RuntimeMigrator.java:83) ~[c7-data-migrator-core-0.1.0-SNAPSHOT.jar!/:0.1.0-SNAPSHOT]
	at io.camunda.migrator.app.MigratorApp.migrateRuntime(MigratorApp.java:58) ~[!/:0.1.0-SNAPSHOT]
	at io.camunda.migrator.app.MigratorApp.main(MigratorApp.java:43) ~[!/:0.1.0-SNAPSHOT]
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103) ~[na:na]
	at java.base/java.lang.reflect.Method.invoke(Method.java:580) ~[na:na]
	at org.springframework.boot.loader.launch.Launcher.launch(Launcher.java:102) ~[c7-data-migrator.jar:0.1.0-SNAPSHOT]
	at org.springframework.boot.loader.launch.Launcher.launch(Launcher.java:64) ~[c7-data-migrator.jar:0.1.0-SNAPSHOT]
	at org.springframework.boot.loader.launch.PropertiesLauncher.main(PropertiesLauncher.java:580) ~[c7-data-migrator.jar:0.1.0-SNAPSHOT]
Caused by: io.camunda.client.api.command.ClientStatusException: io exception
	at io.camunda.client.impl.CamundaClientFutureImpl.transformExecutionException(CamundaClientFutureImpl.java:115) ~[camunda-client-java-8.8.0-SNAPSHOT.jar!/:8.8.0-SNAPSHOT]
	at io.camunda.client.impl.CamundaClientFutureImpl.join(CamundaClientFutureImpl.java:53) ~[camunda-client-java-8.8.0-SNAPSHOT.jar!/:8.8.0-SNAPSHOT]
	at io.camunda.migrator.RuntimeMigrator.lambda$activateMigratorJobs$10(RuntimeMigrator.java:201) ~[c7-data-migrator-core-0.1.0-SNAPSHOT.jar!/:0.1.0-SNAPSHOT]
	at io.camunda.migrator.ExceptionUtils.callApi(ExceptionUtils.java:30) ~[c7-data-migrator-core-0.1.0-SNAPSHOT.jar!/:0.1.0-SNAPSHOT]
	... 10 common frames omitted
```

core/src/test/java/io/camunda/migrator/RuntimeMigratorTest.java
tasso94, line 1: Unfortunately, these mocked tests break easily after the implementation is changed slightly. I attempted to fix them in a time box but ran out of time. 🙁

joaquinfelici: I created https://github.com/camunda/camunda-bpm-platform/issues/5194 to fix this and try to make them less implementation dependent.

joaquinfelici: Now that I actually see the changes, indeed it looks like these tests don't make much sense for now. We're changing implementation everyday, and we already have the qa/ tests to check that migration happens as expected. I suggest we actually remove them for now. What do you think?

core/src/main/java/io/camunda/migrator/ExceptionUtils.java
```diff
@@ -0,0 +1,55 @@
...
+package io.camunda.migrator;
+
+import io.camunda.client.api.command.ClientException;
+import java.util.function.Supplier;
+import org.apache.ibatis.exceptions.PersistenceException;
+import org.camunda.bpm.engine.ProcessEngineException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ExceptionUtils {
```
joaquinfelici, line 17: Now that the class count is increasing, should we start thinking about creating more specific packages like `util` or `exception`?

core/src/main/java/io/camunda/migrator/ExceptionUtils.java
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * Wraps {@link ClientException}, {@link ProcessEngineException}, and {@link PersistenceException} into {@link RuntimeMigratorException}.
+   */
+  public static void callApi(Runnable runnable, String message) {
+    try {
+      runnable.run();
+    } catch (ClientException | ProcessEngineException | PersistenceException e) {
+      var exception = new RuntimeMigratorException(message, e);
+      LOGGER.error(message, exception);
+      throw exception;
```
joaquinfelici, line 51: Since we're refactoring, could be nice to have a method like `wrapInRuntimeException(message, e)` and reuse it above.

core/src/main/java/io/camunda/migrator/Pagination.java
```diff
@@ -0,0 +1,90 @@
...
+  public List<T> toList() {
+    List<T> list = new ArrayList<>();
+    callback(list::add);
+    return list;
+  }
+
+  public <K, U> Map<K, U> toMap(Function<? super T, ? extends K> keyMapper,
+                                Function<? super T, ? extends U> valueMapper) {
+    return toList().stream().collect(Collectors.toMap(keyMapper, valueMapper));
+  }
```
joaquinfelici, line 88: I think it could be really nice to have simple unit tests for this class, especially for methods like `toList()` and `toMap()`.

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
```diff
@@ -131,49 +128,33 @@ protected void storeMapping(String legacyProcessInstanceId, Long processInstance
...
-          .variables(globalVariables)
-          .send()
-          .join()
-          .getProcessInstanceKey();
-    } catch (ProcessEngineException | ClientException e) {
-      LOGGER.error("An error occurred while starting new process instance with legacyId {}, the migration will halt", legacyProcessInstanceId);
-      throw new MigratorException("Error while migrating process instance with legacyId " + legacyProcessInstanceId, e);
-    }
+  protected long startNewProcessInstance(String legacyProcessInstanceId) {
+    var processInstanceQuery = runtimeService.createProcessInstanceQuery().processInstanceId(legacyProcessInstanceId);
```
joaquinfelici, line 139: Should also be wrapped in `callApi()`

joaquinfelici: Same below on `camundaClient.newCreateInstanceCommand()` right? And inside of `getGlobalVariables()`.
Maybe we could wrap this whole method call 💡

joaquinfelici (commented):
Nice changes, it's cool to see api calls, exceptions and pagination more centralized.
Added a few suggestions.

core/src/main/java/io/camunda/migrator/ExceptionUtils.java
tasso94: 👍

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
tasso94: We only need to wrap for methods that throw one of the expected exceptions. `runtimeService.createProcessInstanceQuery().processInstanceId(legacyProcessInstanceId)` is just a fluent builder where none of the expected exceptions are thrown yet. Only when calling `processInstanceQuery::singleResult` can the expected exceptions be thrown, which is wrapped in a `callAPI`. 
Or am I missing something?

tasso94: The `Pagination` class handles `callAPI` itself.

joaquinfelici: Ah indeed makes sense, I missed that you split this call. Looks good to me.

tasso94: Same as above. Only methods that actually executed the query like `listPage` or `singleResult` need to be wrapped. This should be the case here if I'm not missing something? 🤔

core/src/test/java/io/camunda/migrator/RuntimeMigratorTest.java
```diff
@@ -63,13 +62,14 @@ public void shouldMigrationExceptionBeThrownOnPersistenceException() {
     // when migrating, then exception is wrapped and migration halted
     MigratorException migratorException = assertThrows(MigratorException.class, () -> runtimeMigrator.migrate());
     assertEquals(caughtException, migratorException.getCause());
-    assertTrue(migratorException.getMessage().contains("Error while fetching instances to migrate"));
+    assertTrue(migratorException.getMessage().contains("Error occurred: shutting down Data Migrator gracefully."));
```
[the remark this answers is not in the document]
tasso94: Done 👍

PR #51, opened 2025-06-04
feat(runtime): introduce --list-skipped and --retry-skipped

joaquinfelici:
Related to https://github.com/camunda/camunda-bpm-platform/issues/4986

### DIFF
--- a/assembly/resources/README.txt
+++ b/assembly/resources/README.txt
@@ -25,7 +25,7 @@ Options
 - `--runtime` – Migrate runtime data
 - `--history` – Migrate history data
 - `--list-skipped` – List previously skipped data
-- `--retry-skipped` – Retry previously skipped data
+- `--retry-skipped` – Only retry the migration of previously skipped data
 
 ==================
 


core/src/main/resources/logback-spring.xml
joaquinfelici, line 1: This file guarantees that the PRINTER logger uses an independent appender that won't be impacted by the level set in the application.yml and that has a special format (raw output) so that the output is directly consumable by cockpit.

core/src/main/java/io/camunda/migrator/PrintUtils.java
```diff
@@ -0,0 +1,20 @@
...
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.migrator;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PrintUtils {
```
joaquinfelici, line 13: Originally this class contained more logic, that's why I created it. Now that I simplified it, we could remove it and get the printer logger directly, but I find like it's interesting to have it as a separate concern and will make it much easier to refactor if we want to change the implementation.

assembly/resources/README.txt
```diff
@@ -24,6 +24,8 @@ Options
 
 - `--runtime` – Migrate runtime data
 - `--history` – Migrate history data
+- `--list-skipped` – List previously skipped data
+- `--retry-skipped` – Retry previously skipped data
```
HeleneW-dot, line 28: Q: is this in addition to other data or will the migrator only be migrating skipped data with this flag?

qa/src/test/java/io/camunda/migrator/qa/SkipAndRetryProcessInstancesTest.java
```diff
@@ -126,18 +133,61 @@ public void shouldLogWarningWhenProcessInstanceHasBeenCompleted(CapturedOutput o
...
+  @Test
+  public void shouldListSkippedProcessInstances(CapturedOutput output) {
+    // given skipped process instance
+    deployProcessInC7AndC8("multiInstanceProcess.bpmn");
+    List<String> processInstancesIds = new ArrayList<>();
+    for (int i = 0; i < 10; i++) {
+      processInstancesIds.add(runtimeService.startProcessInstanceByKey("multiInstanceProcess").getProcessInstanceId());
+    }
+    runtimeMigrator.start();
+    ensureTrue("Unexpected state: one process instance should be skipped", idKeyMapper.findSkippedProcessInstanceIds().size() == 10);
```
HeleneW-dot, line 159: All 10 are skipped right? Also could we use the count method here?
```suggestion
    ensureTrue("Unexpected state: 10 process instances should be skipped", idKeyMapper. findSkippedProcessInstanceIdsCount() == 10);
```

HeleneW-dot (approved):
Lgtm! just some small questions

HeleneW-dot (commented):
Actually reading the ticket description I have one more question:
```
It should be possible to combine the flags like this:
--runtime --list-skipped
--history --retry-skipped
```

Would it be possible to add a test for the runtime combo (since history is not supported yet anyway)?

qa/src/test/java/io/camunda/migrator/qa/SkipAndRetryProcessInstancesTest.java
joaquinfelici: Indeed! FIxed

joaquinfelici:
@HeleneW-dot regarding [this comment](https://github.com/camunda/c7-data-migrator/pull/51#pullrequestreview-2915985797) (I still don't understand why GH will let us reply some comments but not others). We're not testing script flags at all today.

assembly/resources/README.txt
joaquinfelici: Maybe we could make it clearer?
```suggestion
- `--retry-skipped` – Only retry the migration of previously skipped data
```

HeleneW-dot:
> @HeleneW-dot regarding [this comment](https://github.com/camunda/c7-data-migrator/pull/51#pullrequestreview-2915985797) (I still don't understand why GH will let us reply to some comments but not others). We're not testing script flags at all today.

Gotcha all good then

assembly/resources/start.bat
```diff
@@ -40,11 +40,12 @@ goto :end
...
-echo Usage: start.bat [--runtime] [--history] [--retry]
+echo Usage: start.bat [--runtime] [--history] [--list-skipped|--retry-skipped]
 echo Options:
-echo   --runtime     - Migrate runtime data only
-echo   --history     - Migrate history data only
-echo   --retry       - Retry only previously skipped data
+echo   --runtime          - Migrate runtime data only
+echo   --history          - Migrate history data only
+echo   --list-skipped     - List previously skipped data
+echo   --retry-skipped    - Retry only previously skipped data
```
joaquinfelici, line 48:
```suggestion
echo   --retry-skipped    - Only retry the migration of previously skipped data
```

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
```diff
@@ -86,6 +95,14 @@ public void migrate() {
     activateMigratorJobs();
   }
 
+  protected List<String> listSkippedProcessInstances() {
+   return new Pagination<String>()
+        .batchSize(batchSize)
+        .maxCount(idKeyMapper::findSkippedProcessInstanceIdsCount)
+        .page(offset -> idKeyMapper.findSkippedProcessInstanceIds(0, batchSize))
```
tasso94, line 102: /q When skipped instances > 500 doesn't this always return the same 500 skipped instances?

core/src/main/resources/logback-spring.xml
tasso94, line 1: /q Why do we need to configure this using an XML file?

core/src/main/java/io/camunda/migrator/PrintUtils.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+package io.camunda.migrator;
+
+import java.util.List;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PrintUtils {
+
+  protected static final Logger PRINTER = LoggerFactory.getLogger("PRINTER");
```
tasso94, line 16: /q Why does the name of this logger not follow the usual Java package naming schema?

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
```diff
@@ -86,6 +95,14 @@ public void migrate() {
...
+        .toList();
```
tasso94, line 103: /q When passing a `#callback` function that delegates to the printer, we can avoid loading all IDs into the memory at once. Printing in batches would avoid this.

tasso94 (commented):
Looks good to me overall. However, I have some questions.

core/src/main/resources/logback-spring.xml
joaquinfelici: I wrote something about this [in this comment](https://github.com/camunda/c7-data-migrator/pull/51#discussion_r212659708). I explored other options (like System:out and JUL), but thought this one to make more sense. Let me know what you think.

core/src/main/java/io/camunda/migrator/PrintUtils.java
joaquinfelici: Mentioned this [here](https://github.com/camunda/c7-data-migrator/pull/51#discussion_r2126597081) too. If it follows the convention and they configure io.camunda.migrator* with WARN or ERROR, the output would be muted. We want this to always be printed independently of logger level.

PR #53, opened 2025-06-04
chore(core): reuse sql in count and query

yanavasileva:
https://github.com/camunda/camunda-bpm-platform/issues/4991


tasso94 (approved):
Cool that you figured this out! 👍

PR #58, opened 2025-06-11
chore(docs): document async prerequisite

HeleneW-dot:
related to [#5172](https://github.com/camunda/camunda-bpm-platform/issues/5172)

Document limitation around async wait states

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,8 @@
 1. Navigate to Operate and check result.
 
 #### Migration Limitations
+- Async before/after wait states:
+  - C8 does not support async before/after wait states. Therefore, if you were to migrate an instance currently waiting asynchronously at an element in a C7 model, this instance would just continue without waiting in the equivalent C8 model. Please adjust your model's logic accordingly prior to migration
 - Message events:
   - only message catch and throw events are supported for migration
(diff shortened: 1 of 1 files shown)


HeleneW-dot:
@yanavasileva putting into review but keeping the ticket as there will be more PRs

README.md
```diff
@@ -16,6 +16,8 @@
 1. Navigate to Operate and check result.
 
 #### Migration Limitations
+- Async before/after wait states:
+  - C8 does not support async before/after wait states. Therefore, if you were to migrate an instance currently waiting asynchronously at an element in a C7 model, this instance would just continue without waiting in the equivalent C8 model. Please adjust your model's logic accordingly prior to migration
```
yanavasileva, line 20: 🔧 Use full phrase. Maybe we can even add link to the docs: https://docs.camunda.org/manual/latest/user-guide/process-engine/transactions-in-processes/#asynchronous-continuations
```suggestion
  - C8 does not support Asynchronous continuation (before/after) wait states. Therefore, if you were to migrate an instance currently waiting asynchronously at an element in a C7 model, this instance would just continue without waiting in the equivalent C8 model. Please adjust your model's logic accordingly prior to migration
```

yanavasileva (approved):
👍 Only small suggestion.

PR #60, opened 2025-06-11
chore(assembly): add deployment dir to start.bat

yanavasileva:
https://github.com/camunda/camunda-bpm-platform/issues/5156


tasso94 (approved):
Nice catch!

yanavasileva:
We still are in code freeze, not merging for now.

PR #61, opened 2025-06-12
fix(core): handle variable with `null` values

tasso94:
related to camunda/camunda-bpm-platform#5201


venetrius (approved):
👍 Looks good to me!

PR #63, opened 2025-06-13
fix(qa) SubprocessMigrationTest

venetrius:
related to: camunda/camunda-bpm-platform/issues/5167


yanavasileva (changes requested):
CI fails

venetrius:
The test failed because we override the `legacyId` variable for the call activity with the subprocess `legacyId`. 
Added commit to fix. If it needs more work (PR feedback) maybe I can temporarily comment out the test and merge to fix CI.

yanavasileva:
@venetrius, I will need more time to review latest changes for the legacy variable. Let's disable the test for now (+ fix the compilation error) to fix CI on main sooner.

venetrius:
@yanavasileva 
Reverted changes to the `RuntimeMigrator`.
Will pick reverted commit to new PR after this one is merged.

venetrius:
Created follow up PR https://github.com/camunda/c7-data-migrator/pull/64

PR #72, opened 2025-06-16
ci: daily run + distro upload

yanavasileva:


HeleneW-dot:
The ticket also says `Re-run in case of common failures` but I dont see any changes related to this. Will we address that in a follow up? Are we thinking a similar mechanism like for optimize ci?

yanavasileva:
daily run of release didn't worked out so I will revert that part: https://github.com/camunda/c7-data-migrator/pull/74

PR #77, opened 2025-06-20
chore(validation): add validation for model and element presence

HeleneW-dot:
related to [#5187](https://github.com/camunda/camunda-bpm-platform/issues/5187)

- I created [this follow up ticket](https://github.com/camunda/camunda-bpm-platform/issues/5241) for future improvements and refactoring of validation logic
- For now, I stuck with the per instance validation logic to keep effort low even though that approach may not be the most performant
- I added a cache to reduce querying for the C8 model multiple times when we have multiple instances from the same model. For future iterations it would be neat to make the cache size configurable, we could also adjust the cache to expire based on size rather than amount of definitions. Since the validation may be refactored a bit anyway I did not invest time into that for now but added a note in the follow up ticket

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ However, even though the C7 Data Migrator is not yet ready for production, we en
 1. Deploy migrated C8 process models and any other required resources such as referenced DMN or forms.
 1. Build or download the distribution.
 1. Start Migrator (start.sh/start.bat) and wait to finish.
+   2. Do not make any changes to the C8 deployments while the migrator is running.
 1. Navigate to Operate and check result.
 1. When a process instance gets skipped, you can start C7 again, modify the process instance into a supported state and shut down C7 again.
 
--- a/core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
+++ b/core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
@@ -88,13 +88,11 @@ public class RuntimeMigrator {
   }
 
   public void migrate() {
-    Map<String, Long> c8DefinitionIdToDefinitionKey = getC8DeploymentIdToKeyMap();
-
     fetchProcessInstancesToMigrate(legacyProcessInstance -> {
 
       String legacyProcessInstanceId = legacyProcessInstance.id();
       Date startDate = legacyProcessInstance.startDate();
-      if (skipProcessInstance(legacyProcessInstanceId, c8DefinitionIdToDefinitionKey)) {
+      if (skipProcessInstance(legacyProcessInstanceId)) {
         LOGGER.info("Skipping process instance with legacyId: {}", legacyProcessInstanceId);
         if (!legacyProcessInstance.skippedPreviously()) {
           storeMapping(legacyProcessInstanceId, startDate, null);
@@ -124,9 +122,9 @@ public class RuntimeMigrator {
         .callback(PrintUtils::print);
   }
 
-  protected boolean skipProcessInstance(String legacyProcessInstanceId, Map<String, Long> c8DefinitionIdToDefinitionKey) {
+  protected boolean skipProcessInstance(String legacyProcessInstanceId) {
     try {
-      validateProcessInstanceState(legacyProcessInstanceId, c8DefinitionIdToDefinitionKey);
+      validateProcessInstanceState(legacyProcessInstanceId);
     } catch (IllegalStateException e) {
       LOGGER.warn("Process instance with legacyId [{}] can't be migrated: {}", legacyProcessInstanceId, e.getMessage());
       return true;
@@ -226,10 +224,9 @@ public class RuntimeMigrator {
   /**
    * This method iterates over all the activity instances of the root process instance and its
    * children until it either finds an activityInstance that cannot be migrated or the iteration ends.
-   * For now, only multi-instance activities will fail validation.
    * @param legacyProcessInstanceId the legacy id of the root process instance.
    */
-  protected void validateProcessInstanceState(String legacyProcessInstanceId, Map<String, Long> c8DefinitionIdToDefinitionKey) {
+  protected void validateProcessInstanceState(String legacyProcessInstanceId) {
     LOGGER.debug("Validate legacy process instance by ID: {}", legacyProcessInstanceId);
     ProcessInstanceQuery processInstanceQuery = runtimeService.createProcessInstanceQuery()
         .rootProcessInstanceId(legacyProcessInstanceId);
@@ -242,17 +239,23 @@ public class RuntimeMigrator {
           String c7DefinitionId = processInstance.getProcessDefinitionId();
           String c8DefinitionId = processInstance.getProcessDefinitionKey();
 
-          // validate C8 deployment exists
-          if (!c8DefinitionIdToDefinitionKey.containsKey(c8DefinitionId)) {
+          var c8DefinitionSearchRequest = camundaClient.newProcessDefinitionSearchRequest()
+              .filter(filter -> filter.processDefinitionId(c8DefinitionId))
+              .sort((s) -> s.version().desc());
+
+          List<ProcessDefinition> c8Definitions = callApi(c8DefinitionSearchRequest::execute).items();
+          if (c8Definitions.isEmpty()) {
             throw new IllegalStateException(
                 String.format("No C8 deployment found for process ID [%s] required for instance with legacyID [%s].",
                     c8DefinitionId, legacyProcessInstanceId));
           }
 
           ActivityInstance activityInstanceTree = callApi(() -> runtimeService.getActivityInstance(processInstanceId));
-          BpmnModelInstance c7BpmnModelInstance = callApi(
-              () -> repositoryService.getBpmnModelInstance(c7DefinitionId));
-          String c8XmlString = camundaClient.newProcessDefinitionGetXmlRequest(c8DefinitionIdToDefinitionKey.get(c8DefinitionId)).execute();
+          BpmnModelInstance c7BpmnModelInstance = callApi(() -> repositoryService.getBpmnModelInstance(c7DefinitionId));
+
+          long processDefinitionKey = c8Definitions.getFirst().getProcessDefinitionKey();
+          String c8XmlString =
+              callApi(() -> camundaClient.newProcessDefinitionGetXmlRequest(processDefinitionKey).execute());
           BpmnModelInstance c8BpmnModelInstance = Bpmn.readModelFromStream(new ByteArrayInputStream(c8XmlString.getBytes(StandardCharsets.UTF_8)));
 
           LOGGER.debug("Collecting active descendant activity instances for legacyId [{}]", processInstanceId);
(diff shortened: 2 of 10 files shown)


qa/src/test/java/io/camunda/migrator/qa/RuntimeMigrationAbstractTest.java
```diff
@@ -83,12 +87,20 @@ protected void deployCamunda7Process(String resourcePath) {
...
     if (deployment == null) {
       throw new IllegalStateException("Could not deploy process");
     }
+
+    Awaitility.await().ignoreException(ClientException.class).untilAsserted(() -> {
+      List<ProcessDefinition> c8Deployments = camundaClient.newProcessDefinitionSearchRequest()
+          .filter(filter-> filter.resourceName(resourcePath))
+          .send().join().items();
+      assertThat(c8Deployments).hasSize(1);
+    });
```
HeleneW-dot, line 103: Can remove again once Tassilos changes are merged

qa/src/test/resources/io/camunda/migrator/bpmn/c7/userTaskProcessWithMissingUserTaskInC8.bpmn
```diff
@@ -0,0 +1,45 @@
...
+      <bpmn:incoming>Flow_1ur86ut</bpmn:incoming>
+    </bpmn:endEvent>
+    <bpmn:sequenceFlow id="Flow_1ur86ut" sourceRef="userTaskId" targetRef="Event_0togu4y" />
+    <bpmn:userTask id="userTaskId" name="UserTaskName">
+      <bpmn:incoming>Flow_0xg7od7</bpmn:incoming>
+      <bpmn:outgoing>Flow_1ur86ut</bpmn:outgoing>
+    </bpmn:userTask>
+  </bpmn:process>
+  <bpmndi:BPMNDiagram id="BPMNDiagram_1">
+    <bpmndi:BPMNPlane id="BPMNPlane_1" bpmnElement="userTaskProcessId">
```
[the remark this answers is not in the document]
HeleneW-dot: I think this doesnt matter

qa/src/test/java/io/camunda/migrator/qa/RuntimeMigrationAbstractTest.java
```diff
@@ -28,6 +31,7 @@
 import org.junit.jupiter.api.AfterEach;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
+import org.testcontainers.shaded.org.awaitility.Awaitility;
```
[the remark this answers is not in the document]
HeleneW-dot: These will be removed again once Tassilos changes are merged

tasso94 (changes requested):
Thank you! Looks already quite good to me. 👍

I appreciate your initiative to improve performance, but adding the caching logic at this point may not be the best approach given the current unknowns and potential increase in complexity.

Here is what I think in more detail:
* The caching logic introduces additional complexity to the Minimum Viable Product (MVP).
  * This complexity can make debugging more challenging and increase the maintenance burden going forward.
  * At this point, we should aim for a solution that is as simple as possible.
* Adding a new dependency for caching means we have to manage it by ensuring it is license cleared and kept updated. Tracking any security vulnerabilities or bugs in this third-party dependency adds to our overhead.
* We have not conducted a holistic review to determine the best approach for overall performance improvements.
  * Your implementation optimizes performance for one specific case of the data migrator.
  * A more comprehensive design-thinking process might reveal different and potentially better optimization strategies or reveal other areas that benefit much more from performance improvements.
  * We have yet to measure the impact of the current unoptimized approach.
Without performance benchmarks or real user feedback, we might be adding unnecessary complexity.

My proposal would be:
* Simplify your pull request by removing the caching logic.
* Let's create a follow-up task that captures your idea(s) to optimize the validation logic.
* Based on measuring the current performance and user feedback we can prioritize investments into performance improvements.
  * If performance comes up as an issue in the future, I would probably plan a project to review and identify where our efforts can yield the most significant performance improvements with the highest impact.

Thank you!

HeleneW-dot:
> * Let's create a follow-up task that captures your idea(s) to optimize the validation logic.

The follow up task is already linked in the description :)

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
```diff
@@ -233,21 +239,41 @@ protected void validateProcessInstanceState(String legacyProcessInstanceId) {
...
-          ActivityInstance activityInstanceTree = callApi(() -> runtimeService.getActivityInstance(processInstanceId));
+          String c7DefinitionId = processInstance.getProcessDefinitionId();
+          String c8DefinitionId = processInstance.getProcessDefinitionKey();
+
+          // validate C8 deployment exists
+          if (!c8DefinitionIdToDefinitionKey.containsKey(c8DefinitionId)) {
+            throw new IllegalStateException(
+                String.format("No C8 deployment found for process ID [%s] required for instance with legacyID [%s].",
+                    c8DefinitionId, legacyProcessInstanceId));
+          }
```
tasso94, line 250:
```suggestion
          var searchRequest = camundaClient.newProcessDefinitionSearchRequest()
              .filter(filter -> filter.processDefinitionId(c8DefinitionId))
              .sort((s) -> s.version().desc());

          List<ProcessDefinition> processDefinitions = callApi(() -> searchRequest.send().join()).items();
          if (processDefinitions.isEmpty()) { // empty means process not deployed on C8.
            throw new IllegalStateException(
                String.format("No C8 deployment found for process ID [%s] required for instance with legacyID [%s].",
                    c8DefinitionId, legacyProcessInstanceId));
          }
```

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
```diff
@@ -233,21 +239,41 @@ protected void validateProcessInstanceState(String legacyProcessInstanceId) {
...
 
-          String processDefinitionId = processInstance.getProcessDefinitionId();
-          BpmnModelInstance bpmnModelInstance = callApi(() -> repositoryService.getBpmnModelInstance(processDefinitionId));
+          ActivityInstance activityInstanceTree = callApi(() -> runtimeService.getActivityInstance(processInstanceId));
+          BpmnModelInstance c7BpmnModelInstance = callApi(
+              () -> repositoryService.getBpmnModelInstance(c7DefinitionId));
+          String c8XmlString = camundaClient.newProcessDefinitionGetXmlRequest(c8DefinitionIdToDefinitionKey.get(c8DefinitionId)).execute();
```
tasso94, line 255:
```suggestion
          BpmnModelInstance c7BpmnModelInstance = callApi(() -> repositoryService.getBpmnModelInstance(c7DefinitionId));

          long processDefinitionKey = processDefinitions.getFirst().getProcessDefinitionKey();
          String c8XmlString = camundaClient.newProcessDefinitionGetXmlRequest(processDefinitionKey).execute();
```

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
```diff
@@ -339,4 +365,25 @@ public void setMode(MigratorMode mode) {
...
+      }).execute().items();
+
+      idToKey.putAll(c8Deployments.stream()
+          .collect(
+              Collectors.toMap(ProcessDefinition::getProcessDefinitionId, ProcessDefinition::getProcessDefinitionKey)));
+      from = from + c8Deployments.size();
+    } while (c8Deployments.size() == batchSize);
+
+    return idToKey;
+  }
```
tasso94, line 387: /suggest
Instead of querying for all process definitions all the time, can we just query for the process definition that we actually need on validation? There may be more process definitions deployed on C8 than on C7, but we only need the C8 process definition key for C7 legacy process instances.

See code suggestions above.

tasso94 (changes requested):

tasso94 (changes requested):
Looks already very good now. However, I have some more questions.

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
```diff
@@ -233,21 +236,47 @@ protected void validateProcessInstanceState(String legacyProcessInstanceId) {
...
         .callback(processInstance -> {
           String processInstanceId = processInstance.getId();
+          String c7DefinitionId = processInstance.getProcessDefinitionId();
+          String c8DefinitionId = processInstance.getProcessDefinitionKey();
+
+          var c8DefinitionSearchRequest = camundaClient.newProcessDefinitionSearchRequest()
+              .filter(filter -> filter.processDefinitionId(c8DefinitionId))
+              .sort((s) -> s.version().desc());
+
+          List<ProcessDefinition> c8Definitions = callApi(c8DefinitionSearchRequest::execute).items();
```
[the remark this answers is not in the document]
HeleneW-dot: Yes

qa/src/test/java/io/camunda/migrator/qa/ProcessDefinitionNotFoundTest.java
```diff
@@ -11,39 +11,73 @@
...
+    logs.getEvents().clear();
+
+    // when
+    runtimeMigrator.start();
+
+    // then
+    logs.assertContains(String.format(
+        "Process instance with legacyId [%s] can't be migrated: "
+            + "No C8 deployment found for process ID [%s] required for instance with "
+            + "legacyID [%s].",  c7Instance.getId(), "simpleProcess", c7Instance.getId()));
```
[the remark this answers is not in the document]
HeleneW-dot: Ah, I just replaced the exception assertion with a log assertion without evaluating the test logic closer. Will have a look

qa/pom.xml
```diff
@@ -42,6 +42,24 @@
...
+    <artifactId>logunit-logback</artifactId>
+    <version>2.0.0</version>
+    <exclusions>
+      <exclusion>
+        <groupId>ch.qos.logback</groupId>
+        <artifactId>logback-classic</artifactId>
+      </exclusion>
+    </exclusions>
+    <scope>test</scope>
+  </dependency>
```
tasso94, line 62:
```suggestion
    <dependency>
      <groupId>io.github.netmikey.logunit</groupId>
      <artifactId>logunit-logback</artifactId>
      <version>2.0.0</version>
      <exclusions>
        <exclusion>
          <groupId>ch.qos.logback</groupId>
          <artifactId>logback-classic</artifactId>
        </exclusion>
      </exclusions>
      <scope>test</scope>
    </dependency>
```

qa/src/test/java/io/camunda/migrator/qa/SkipAndRetryProcessInstancesTest.java
```diff
@@ -15,6 +15,8 @@
 import static org.camunda.bpm.engine.impl.util.EnsureUtil.ensureTrue;
 
 import io.camunda.client.api.search.response.ProcessInstance;
+import io.camunda.migrator.RuntimeMigrator;
+import io.github.netmikey.logunit.api.LogCapturer;
```
tasso94, line 19:
```suggestion
```

qa/src/test/java/io/camunda/migrator/qa/SkipAndRetryProcessInstancesTest.java
```diff
@@ -26,6 +28,7 @@
 import org.camunda.bpm.engine.task.Task;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.extension.RegisterExtension;
```
tasso94, line 31:
```suggestion
```

README.md
```diff
@@ -18,6 +18,7 @@ However, even though the C7 Data Migrator is not yet ready for production, we en
 1. Deploy migrated C8 process models and any other required resources such as referenced DMN or forms.
 1. Build or download the distribution.
 1. Start Migrator (start.sh/start.bat) and wait to finish.
+   2. Do not make any changes to the C8 deployments while the migrator is running.
```
tasso94, line 21: Is this still a problem after the refactoring? 🤔

tasso94 (approved):
Looks good to me. 👍 
Just some more smaller hints.

PR #79, opened 2025-06-23
fix(core): skip deserialization of variables

yanavasileva:
https://github.com/camunda/camunda-bpm-platform/issues/5174

### DIFF
--- a/qa/src/test/java/io/camunda/migrator/qa/BatchConfigurationTest.java
+++ b/qa/src/test/java/io/camunda/migrator/qa/BatchConfigurationTest.java
@@ -13,7 +13,6 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import io.camunda.client.api.command.ClientException;
 import io.camunda.client.api.search.response.ProcessDefinition;
 import java.util.List;
-import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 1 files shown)


yanavasileva:
Not ready for review, I have a bug in the test.

core/src/main/java/io/camunda/migrator/Pagination.java
```diff
@@ -89,8 +95,17 @@ public List<T> toList() {
   public Map<String, Object> toVariableMap() {
     Map<String, Object> result = new HashMap<>();
     toList().forEach(e -> {
-      VariableInstance var = (VariableInstance) e;
-      result.put(var.getName(), var.getValue());
+      VariableInstanceEntity var = (VariableInstanceEntity) e;
+      TypedValue typedValue = var.getTypedValue(false);
+      if (typedValue.getType().equals(ValueType.OBJECT)) {
+        // don't serialize the value for object
```
venetrius, line 101: 🙃  I think this is skipping `deserialization` and not `serialization`

core/src/main/java/io/camunda/migrator/Pagination.java
```diff
@@ -89,8 +95,17 @@ public List<T> toList() {
...
+        result.put(var.getName(), typedValue.getValue());
```
venetrius, line 102: ❓  We don't need to call to `toString` in this case, or is that the default behaviour?

venetrius (approved):
👍 Looks good to me
I had a question, but only because I am not that experienced with values / typed values and not because I think a change is required.

PR #81, opened 2025-06-25
Clarified async limitation - as service task async before works properly

berndruecker:
https://github.com/camunda/camunda-bpm-platform/issues/4992


yanavasileva (approved):
Thank you for your contribution.

PR #82, opened 2025-06-25
feat(core): add variable interceptor

yanavasileva:
https://github.com/camunda/camunda-bpm-platform/issues/4989


yanavasileva:
Todo: 

- [ ] add paragraph in the readme after initial review feedback

core/src/main/java/io/camunda/migrator/Pagination.java
```diff
@@ -88,25 +96,58 @@ public List<T> toList() {
...
-        result.put(var.getName(), typedValue.getValue().toString());
-      } else {
-        result.put(var.getName(), var.getValue());
+      VariableInvocation variableInvocation = new VariableInvocation((VariableInstanceEntity) e);
+      if (interceptors != null && !interceptors.isEmpty()) {
+        interceptors.forEach(i -> {
+          try {
+            i.execute(variableInvocation);
+          } catch (Exception ex) {
+            throw new VariableInterceptorException("An error occurred during variable transformation.", ex);
```
venetrius, line 132: 💭 Maybe we could mention in the README that: 
if an interceptors throws an error no other interceptors will be called for that process instance. Mention that right now we don't give a tool to specify interceptor order.

core/src/main/java/io/camunda/migrator/Pagination.java
```diff
@@ -88,25 +96,58 @@ public List<T> toList() {
     return list;
   }
 
+  public Map<String, Map<String, Object>> toVariableMap() {
```
venetrius, line 99: 🙃  This could be simplified like this: 
```
public Map<String, Map<String, Object>> toVariableMap() {
    Map<String, Map<String, Object>> result = new HashMap<>();
    processVariables((variable, typedValue) -> {
      result
          .computeIfAbsent(variable.getActivityInstanceId(), k -> new HashMap<>())
          .putAll(getVariableResult(variable, typedValue));
    });
    return result;
```

venetrius (changes requested):
Please see my comments.  
The Readme should be updated to.

core/src/main/java/io/camunda/migrator/Pagination.java
```diff
@@ -88,25 +96,58 @@ public List<T> toList() {
...
+  private Map<String, Object> getVariableResult(VariableInstanceEntity variable, TypedValue typedValue) {
+    Map<String, Object> variableResult = new HashMap<>();
+    if (typedValue.getType().equals(ValueType.OBJECT)) {
+      variableResult.put(variable.getName(), typedValue.getValue());
+    } else if (typedValue.getType().equals(SpinValueType.JSON) || typedValue.getType().equals(SpinValueType.XML)) {
+      variableResult.put(variable.getName(), typedValue.getValue().toString());
+    } else {
+      variableResult.put(variable.getName(), variable.getValue());
+    }
+    return variableResult;
```
yanavasileva, line 151: The idea here is to do this in a default interceptor with order(0) and the rest will be executed afterwards. The last in line takes precedence. 
WIP: https://github.com/camunda/c7-data-migrator/pull/100/commits/4be96431ef8ad324534711197c48d1ec786f1dd4
I need to do more validation to confirm the approach.

core/src/main/java/io/camunda/migrator/interceptor/VariableInvocation.java
```diff
@@ -14,40 +14,45 @@
...
-   * Returns the process instance ID associated with this variable invocation.
+   * Returns the DTO representation of the variable.
    *
-   * @return the process instance ID
+   * @return the {@link MigrationVariableDto}
    */
-  public String getProcessInstanceId() {
-    return processInstanceId;
+  public MigrationVariableDto getMigrationVariable() {
+    if (migrationVariable == null) {
```
venetrius, line 43: 🙃  Is it possible for `migrationVariable` to be null? There is only 1 constructor and that sets `migrationVariable`

README.md
```diff
@@ -240,6 +240,54 @@ logging:
...
+2. Annotate your class with `@Component` to register it as a Spring bean.
+3. Override the `execute(VariableInvocation invocation)` method to add your logic.
+
+**Example:**
+```java
+import org.springframework.core.annotation.Order;
+import org.springframework.stereotype.Component;
+
+@Component
+@Order(100) // Optional: controls execution order if multiple interceptors exist
```
venetrius, line 259: 👍  Cool!

venetrius (approved):
👍 Looks good to me! Good job!
I had a nitpick. 

An other thought 💭 : Right now the customer can access the variable name and variable value in a custom interceptor. I can imagine that in the future we get an ask to also include the process definition name, activity name or C7 datatype etc. and I think the current solution can be extended easily in that case.

README.md
```diff
@@ -240,6 +240,54 @@ logging:
 | | `.level.io.camunda.migrator` | `string`  | Migrator logging level. Default: `INFO`                                                                                                   |
 | | `.file.name`                 | `string`  | Log file location. Set to: `logs/c7-data-migrator.log`. If not specified, logs are output to the console.                                 |
 
+## Variable transformation
```
tasso94, line 243: /suggest I would suggest describing high-level how to build an extension JAR and where to drop it.

yanavasileva: Will follow up in https://github.com/camunda/c7-data-migrator/pull/113

PR #94, opened 2025-06-30
feat(test): extends tests for oracle and postgresql

joaquinfelici:
Related to https://github.com/camunda/camunda-bpm-platform/issues/5207

```shell
mvn verify               # run tests with h2
mvn verify -Poracle      # run tests with oracle
mvn verify -Ppostgresql  # run tests with postgresql
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,8 +35,14 @@ jobs:
           vault-role-id: ${{ secrets.VAULT_ROLE_ID }}
           vault-secret-id: ${{ secrets.VAULT_SECRET_ID }}
 
-      - name: Build with Maven
+      - name: Build and run default tests
         run: mvn clean install --batch-mode
+
+      - name: Run tests with PostgreSQL
+        run: mvn test -Dpostgresql
+
+      - name: Run tests with Oracle
+        run: mvn test -Doracle
 
       - name: Collect distro
         id: distro
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 60
     steps:
       - name: Checkout code
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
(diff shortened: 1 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -35,9 +35,15 @@ jobs:
...
-      - name: Build with Maven
+      - name: Build and run default tests
         run: mvn clean install --batch-mode
 
+      - name: Run tests with PostgreSQL
+        run: mvn verify -Ppostgresql
+
+      - name: Run tests with Oracle
+        run: mvn verify -Poracle
+
```
tasso94, line 46: /suggest I would recommend running PostgreSQL and Oracle tests in dedicated jobs. Here is why:
* GitHub runners might be more likely to be killed preemptively the longer they run (60 minutes is IMO long).
* Dedicated jobs can run in parallel, increasing the time to merge PRs (15 minutes vs. 60 minutes)

qa/src/test/java/io/camunda/migrator/qa/MultiDbExtension.java
```diff
@@ -0,0 +1,80 @@
...
+        .withUsername("camunda")
+        .withPassword("camunda")
+        .withReuse(true)
+        .withExposedPorts(5432);
+    postgres.setPortBindings(List.of(POSTGRESQL_PORT + ":5432"));
+    return postgres;
+  }
+
+  protected static OracleContainer createOracleContainer() {
+    OracleContainer oracle = new OracleContainer("gvenzl/oracle-free:latest")
```
tasso94, line 57: /change Can we pin Oracle 23 here to avoid testing unsupported databases going forward?

.github/workflows/ci.yml
joaquinfelici: Does this mean that we'll need to recompile the artifacts for each job?

.github/workflows/ci.yml
```diff
@@ -35,9 +35,15 @@ jobs:
           vault-role-id: ${{ secrets.VAULT_ROLE_ID }}
           vault-secret-id: ${{ secrets.VAULT_SECRET_ID }}
 
-      - name: Build with Maven
+      - name: Build and run default tests
         run: mvn clean install --batch-mode
 
+      - name: Run tests with PostgreSQL
```
venetrius, line 41: ❌  These can be removed from here now that new jobs are introduced

venetrius (changes requested):
👍 Looks good to me! 
❌  The test with Oracle and PSQL is still left in the default build, that should be removed. (maybe it is there now for testing) 
💭  We could also mention in the Readme that  PSQL and Oracle are supported / tested

venetrius (approved):
👍  Looks good to me!
The CI also runs under 20 min. Nice!

PR #105, opened 2025-07-04
chore(release): undo pin dependencies for release alpha-3

joaquinfelici:
Related to https://github.com/camunda/camunda-bpm-platform/issues/5210

⚠️ To be merged after release

### DIFF
--- a/core/src/main/java/io/camunda/migrator/config/mybatis/C8Configuration.java
+++ b/core/src/main/java/io/camunda/migrator/config/mybatis/C8Configuration.java
@@ -388,7 +388,8 @@ public class C8Configuration extends AbstractConfiguration {
         mappingReader,
         batchOperationReader,
         sequenceFlowReader,
-        batchOperationItemReader);
+        batchOperationItemReader,
+        null);
   }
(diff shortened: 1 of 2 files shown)


joaquinfelici:
@yanavasileva rebased ✅

core/src/main/java/io/camunda/migrator/config/mybatis/C8Configuration.java
```diff
@@ -388,7 +388,8 @@ public RdbmsService rdbmsService(
         mappingReader,
         batchOperationReader,
         sequenceFlowReader,
-        batchOperationItemReader);
+        batchOperationItemReader,
+        null);
```
yanavasileva, line 392: ❓ I see that with the new approach we add beans for all fields. Should we do it for this one too?
Example: https://github.com/camunda/c7-data-migrator/blob/a9ed6b5061cfe273c8c3ec6404dc0cca131d0fc3/core/src/main/java/io/camunda/migrator/config/mybatis/C8Configuration.java#L295-L315

PR #107, opened 2025-07-07
fix(core): exclude slf4j transitively pulled in by spin

tasso94:
related to camunda/camunda-bpm-platform#5272


yanavasileva (approved):
👍 
CI Failure is also present in main.

PR #110, opened 2025-07-07
fix(tests): set log level to WARN for concerned tests to avoid flakiness

joaquinfelici:


yanavasileva (approved):
Looks good. There are still a couple of INFO logs that slip through but they does not seem crucial.

PR #111, opened 2025-07-07
chore(IT): add date timer tests

HeleneW-dot:
related to camunda/camunda-bpm-platform/issues/5173


venetrius (approved):
👍 Looks good to me! 
Let's not forget to link this PR in the related issue. Also keep in mind that code freeze is still in effect.

PR #117, opened 2025-07-09
Update README.md

berndruecker:
I wanted to point out you need the EL which is important - and actually not a limitation - but a step in the todos for me


tasso94 (approved):
Looks good to me. Thank you!

PR #118, opened 2025-07-09
fix(core): avoid re-validating known process instance

tasso94:
related to camunda/camunda-bpm-platform#5280

### DIFF
--- a/core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
+++ b/core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
@@ -81,7 +81,7 @@ public class RuntimeMigrator {
   protected MigratorProperties migratorProperties;
 
   @Autowired
-  private ApplicationContext context;
+  protected ApplicationContext context;
 
 
   protected MigratorMode mode = MIGRATE;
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
```diff
@@ -162,11 +162,10 @@ protected void fetchProcessInstancesToMigrate(Consumer<IdKeyDbModel> storeMappin
 
     } else {
       LOGGER.debug("Fetching latest start date of process instances");
-      Date latestStartDate = callApi(() -> idKeyMapper.findLatestStartDateByType(TYPE.RUNTIME_PROCESS_INSTANCE));
-      LOGGER.debug("Latest start date: {}", latestStartDate);
+      IdKeyDbModel latestProcessInstance = callApi(() -> idKeyMapper.findByTypeOrderedByStartDateDesc(TYPE.RUNTIME_PROCESS_INSTANCE));
+      LOGGER.debug("Latest process instance [{}]", latestProcessInstance);
```
yanavasileva, line 166: On the first run this message is `Latest process instance [null]` but that's expected.

yanavasileva (approved):
👍 Looks good.

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
tasso94: Yes. That's was the case already before. This basically means all process instances will be queried.

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
```diff
@@ -99,13 +99,15 @@ public void migrate() {
     fetchProcessInstancesToMigrate(legacyProcessInstance -> {
 
       String legacyProcessInstanceId = legacyProcessInstance.id();
+      boolean unknown = callApi(() -> !idKeyMapper.checkExists(legacyProcessInstanceId));
+
       Date startDate = legacyProcessInstance.startDate();
-      if (skipProcessInstance(legacyProcessInstanceId)) {
+      if (unknown && skipProcessInstance(legacyProcessInstanceId)) {
```
yanavasileva, line 105: test case failure here? SkipAndRetryProcessInstancesTest.shouldSkipAgainAProcessInstanceThatWasSkipped

tasso94:
Hi @yanavasileva,
I think the test failure indicates a problem with the logic or the chosen approach. Let me debug this further.

tasso94:
Hi @yanavasileva, I came up with a different solution. This should work now and cover all cases.

yanavasileva (approved):
👍 Nice refactoring.

PR #119, opened 2025-07-09
feat(history): Initial test setup for history migration

joaquinfelici:
Related to https://github.com/camunda/camunda-bpm-platform/issues/5262


qa/src/test/java/io/camunda/migrator/qa/util/MultiDbExtension.java
```diff
@@ -48,6 +48,7 @@ protected static PostgreSQLContainer<?> createPostgreSQLContainer() {
         .withUsername("camunda")
         .withPassword("camunda")
         .withReuse(true)
+        .withCommand("postgres", "-c", "max_connections=200")
```
joaquinfelici, line 51: The postgresql tests were randomly failing because there were too many client connections. I added this as a temporary fix and I'll create a follow-up ticket to do a more clean fix, like reducing the size of the Hikari pool which increase with the number of tests.

qa/src/test/resources/application.yml
```diff
@@ -0,0 +1,17 @@
...
+        username: sa
+        password: sa
+        driver-class-name: org.h2.Driver
+    c8:
+      data-source:
+        auto-ddl: true
+        jdbc-url: jdbc:h2:mem:migrator
+        username: sa
+        password: sa
+        driver-class-name: org.h2.Driver
```
yanavasileva, line 17: 🔧 C8 data source was removed on purpose for Runtime. If we need it for history,  I suggest creating a profile properties file used in the history tests. Same goes for PG and ORCL. If I am not mistaken, it will reduce the connections and thus the resources. Feel free to challenge.

qa/src/test/java/io/camunda/migrator/qa/history/HistoryTaskMigrationTest.java
```diff
@@ -0,0 +1,51 @@
...
+import io.camunda.search.entities.UserTaskEntity;
+import java.util.List;
+import org.camunda.bpm.engine.task.Task;
+import org.junit.jupiter.api.Test;
+
+public class HistoryTaskMigrationTest extends HistoryMigrationAbstractTest {
+
+  @Test
+  public void shouldHistoricUserTasksBeMigrated() {
+    deployer.deployCamunda7Process("io/camunda/migrator/bpmn/c7/userTaskProcess.bpmn");
```
yanavasileva, line 24: 💭 There's no C8 deployment here. If want to do a deployment, we will need the C8 client that will slow down the test execution.
My concern is: if we don't deploy the C8 process, will the data be consistent on the C8 side?
I think the answer is the C8 deployments are not like C7 deployment; there's no Deployment table and entity on C8 side. So in theory, it should work out-of-the-box.
Further, we will not do validation as we do for the Runtime.

qa/src/test/java/io/camunda/migrator/qa/history/HistoryMigrationAbstractTest.java
```diff
@@ -0,0 +1,119 @@
...
+  @Autowired
+  protected TaskService taskService;
+
+  @Autowired
+  protected RepositoryService repositoryService;
+
+  // C8 ---------------------------------------
+
+  @Autowired
+  protected CamundaClient camundaClient;
```
yanavasileva, line 65: 🔧  This is not used. Let's remove it now and add it only if need it, also the CPT annotation is not defined.

qa/src/test/resources/application.yml
```diff
@@ -0,0 +1,17 @@
+camunda:
```
yanavasileva, line 1: 🔧 If we add this to the code base should/can we remove `qa/src/main/resources/application.yml`. It will be better to consolidate them at one place.

qa/src/test/java/io/camunda/migrator/qa/history/HistoryTaskMigrationTest.java
```diff
@@ -0,0 +1,51 @@
...
+import static io.camunda.search.entities.FlowNodeInstanceEntity.FlowNodeType.START_EVENT;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.camunda.search.entities.ProcessInstanceEntity;
+import io.camunda.search.entities.UserTaskEntity;
+import java.util.List;
+import org.camunda.bpm.engine.task.Task;
+import org.junit.jupiter.api.Test;
+
+public class HistoryTaskMigrationTest extends HistoryMigrationAbstractTest {
```
yanavasileva, line 20: ❓ I added a second test and it failed. Is it something wrong with the setup?

yanavasileva (commented):
👍 Looks good, left questions and small suggestions.

qa/src/test/resources/application.yml
yanavasileva: I thought extending profiles is possible out of the box in Spring Boot but I couldn't find how. Your solution looks good so let's merge and iterate if need later.

yanavasileva (approved):
👍 Looks good. Thank you for considering my feedback.

PR #120, opened 2025-07-10
feat(core): handle externally started process instances

tasso94:
Add support for separate job types for validation and activation, enabling
graceful handling of externally started process instances alongside migrated ones.

Key changes:
- Add `job-type` and `validation-job-type` configuration properties
- Support FEEL expressions in `validation-job-type` for dynamic job type resolution
- Add `camunda.migrator.validation-job-type=DISABLED` option to skip job type validation entirely
- Handle externally started process instances (without legacyId) by skipping migration

This enables mixed environments where both migrated and externally started
process instances can coexist, with externally started instances generating
"noop" jobs that can be handled by separate job workers.

related to https://github.com/camunda/camunda-bpm-platform/issues/5282

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ camunda.client:
 camunda.migrator:
   batch-size: 500                      # Number of records to process in each batch
   job-type: migrator                   # Job type for actual job activation (used for validation and activation unless validation-job-type is defined)
-  validation-job-type: '=if legacyId then "migrator" else "noop"'        # Job type for validation (optional - falls back to job-type if not defined)
+  validation-job-type: '=if legacyId != null then "migrator" else "noop"'        # Job type for validation (optional - falls back to job-type if not defined)
   auto-ddl: true                       # Automatically create/update database schema
   table-prefix: MY_PREFIX_             # Optional table prefix for migrator schema
   data-source: C7                      # Choose if the migrator schema is created on the data source of 'C7' or 'C8'
@@ -195,12 +195,22 @@ camunda.migrator:
 
 ### Job Type Configuration
 
+During migration, the Data Migrator starts new C8 process instances and sets a special `legacyId` variable to link them to their original C7 process instances. The migrator uses execution listeners on start events for its internal migration logic.
+
+However, if users manually start new C8 process instances on models that still have these migration execution listeners, those instances won't have the `legacyId` variable. This creates a problem:
+- The migrator would try to migrate process instances that don't need migration
+- This could cause errors or unexpected behavior
+
+The `validation-job-type` feature solves this by allowing you to use a FEEL expression creating different job types for externally started process instances versus process instances started by the migrator, ensuring only instances that truly need migration are processed by the migrator.
+
 The migrator supports two job type configurations with fallback behavior:
 
 1. **`job-type`**: Used for actual job activation
    - This is the primary job type used when activating jobs in Camunda 8.
    - It is required for the migrator to function correctly.
 2. **`validation-job-type`**: Used for validation purposes (optional)
+    - You can define a FEEL expression that provides different job types based on the process instance context.
+    - The BPMN process will be validated to contain a start event execution listener with the respectively defined FEEL expression.
 
 **Default Behavior:**
 When `validation-job-type` is not defined, `job-type` is used for both validation and activation.
@@ -215,7 +225,7 @@ camunda.migrator:
 ```yaml
 camunda.migrator:
   job-type: migrator                                           # Used for activation
-  validation-job-type: '=if legacyId then "migrator" else "noop"'  # Used for validation with FEEL expression
+  validation-job-type: '=if legacyId != null then "migrator" else "noop"'  # Used for validation with FEEL expression
 ```
 
 **Important Notes:**
@@ -224,7 +234,7 @@ camunda.migrator:
 - The `validation-job-type` can be a FEEL expression (starts with `=`)
 - Set `validation-job-type` to `DISABLED` to disable job type validation entirely
 - Use FEEL expressions only for validation, not for job activation since on job activation, the FEEL expression is already evaluated to a static value
-- Variables referenced in FEEL expressions must exist in the process instance context
+- You can use variables like the special variable `legacyId` in FEEL expression
 
 **FEEL Expression Requirements:**
 When using FEEL expressions in the `validation-job-type` property, you must also specify the same expression in the execution listener of your BPMN process start events. The migrator configuration alone is not sufficient.
@@ -234,7 +244,7 @@ Example BPMN configuration for FEEL expression validation:
 <bpmn:startEvent id="StartEvent_1">
   <bpmn:extensionElements>
     <zeebe:executionListeners>
-      <zeebe:executionListener eventType="end" type="=if legacyId then &quot;migrator&quot; else &quot;noop&quot;" />
+      <zeebe:executionListener eventType="end" type="=if legacyId != null then &quot;migrator&quot; else &quot;noop&quot;" />
     </zeebe:executionListeners>
   </bpmn:extensionElements>
 </bpmn:startEvent>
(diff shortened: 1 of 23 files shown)


yanavasileva:
I haven't started the review on this.

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
```diff
@@ -387,53 +394,63 @@ protected void activateMigratorJobs() {
...
 
-      migratorJobs.forEach(activatedJob -> {
-        String fetchLegacyIdErrorMessage =
-            String.format("Error while fetching legacyId for job with key:" + activatedJob.getProcessInstanceKey());
-        String legacyId = (String) callApi(() -> activatedJob.getVariable("legacyId"), fetchLegacyIdErrorMessage);
-        long processInstanceKey = activatedJob.getProcessInstanceKey();
+      migratorJobs.forEach(job -> {
+        String checkExternallyStartedErrorMessage =
+            "Error while checking if process instance was externally started for job with key: " + job.getProcessInstanceKey();
+        boolean externallyStarted = callApi(() -> !job.getVariables().contains("legacyId"), checkExternallyStartedErrorMessage);
```
HeleneW-dot, line 408: ❓ I may be mistaken here, but this is what I understood so far:
- any variable referenced in the feel expression must exist
- the feel expression will probably reference the `legacyId` variable
- hence, the externally started instances should also have the `legacyId` variable, but probably have it set to an empty string or null

In that case, would `!job.getVariables().contains("legacyId")` not also evaluate to false for externally started instances, as they will contain the variable (although it may have an empty value)?

core/src/test/java/io/camunda/migrator/JobTypeConfigurationTest.java
```diff
@@ -0,0 +1,206 @@
...
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.TestPropertySource;
+
+class JobTypeConfigurationTest {
+
+  @Nested
+  @SpringBootTest
+  class DefaultJobTypeWithoutConfigTest {
```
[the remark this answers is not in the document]
HeleneW-dot: Nvm, read the rest of the tests and found this is already covered in `ValidationNotDisabledTest` 👍

qa/src/test/java/io/camunda/migrator/qa/CustomJobTypeValidationTest.java
```diff
@@ -0,0 +1,77 @@
...
+                + "\\[Event_1px2j50\\] in C8 process with key \\[(\\d+)\\]\\.", id))))
+        .hasSize(1);
+  }
+
+  @Test
+  public void shouldUseCustomJobTypeInListenerNotFoundMessage() {
+    // given
+    deployProcessInC7AndC8("migratorListenerCustomType.bpmn");
+
+    String id = runtimeService.startProcessInstanceByKey("migratorListenerCustomType").getId();
```
HeleneW-dot, line 58: 💭 I did not immediately understand this test case because it looked so similar to the above. Looking at the models I understand: case 1 is that there is no listener at all, and case 2 is that there is a listener, but its got the wrong type right?

Maybe we can adjust model names to make this more obvious:
```suggestion
    deployProcessInC7AndC8("migratorListenerWrongType.bpmn");

    String id = runtimeService.startProcessInstanceByKey("migratorListenerWrongType").getId();
```

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
```diff
@@ -387,53 +394,63 @@ protected void activateMigratorJobs() {
...
+            modifyProcessInstance.activateElement(activityId).withVariables(localVariables, activityId);
+          });
+
+          String batchUpdatedActivitiesErrorMessage = "Error while activating jobs";
+          callApi(
+              () -> ((ModifyProcessInstanceCommandStep1.ModifyProcessInstanceCommandStep3) modifyProcessInstance).execute(),
+              batchUpdatedActivitiesErrorMessage);
+          // no need to complete the job since the modification canceled the migrator job in the start event
+        } else {
+          LOGGER.info("Process instance with key [{}] was externally started, skipping migrator job activation.", job.getProcessInstanceKey());
```
HeleneW-dot, line 452: ❓ Is C8 process instance key the same as what we would call process instance ID in C7?

qa/src/test/java/io/camunda/migrator/qa/CustomJobTypeValidationTest.java
```diff
@@ -0,0 +1,77 @@
...
+    // then
+    assertThatProcessInstanceCountIsEqualTo(0);
+
+    var events = logs.getEvents();
+    assertThat(events.stream()
+        .filter(event -> event.getMessage()
+            .matches(String.format(".*Skipping process instance with legacyId \\[%s\\]: "
+                + "Couldn't find execution listener of type 'custom-job-type' on start event "
+                + "\\[Event_1px2j50\\] in C8 process with key \\[(\\d+)\\]\\.", id))))
+        .hasSize(1);
```
HeleneW-dot, line 50: Nit
So `MigratorListenerNotFoundTest` covers the actual skipping logic and these tests are just there to ensure the error message is parsed correctly with custom configs right? In that case I think we could just use `logs.assertContains()` instead of this custom matching and asserting
(same suggestion also for the other test classes)

qa/src/test/java/io/camunda/migrator/qa/ValidationJobTypeOnlyTest.java
```diff
@@ -0,0 +1,95 @@
...
+import io.camunda.migrator.qa.util.RuntimeMigrationAbstractTest;
+import io.github.netmikey.logunit.api.LogCapturer;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+import org.springframework.test.context.TestPropertySource;
+
+@TestPropertySource(properties = {
+    "camunda.migrator.validation-job-type==if legacyId then \"migrator\" else \"noop\""
+})
+public class ValidationJobTypeOnlyTest extends RuntimeMigrationAbstractTest {
```
HeleneW-dot, line 23: 💭 Personally, because we already have the extensive config check that confirms when the validation is enabled/disabled and how the job type and validation job type properties are set, I feel like we dont need this class to confirm the behaviour when only the validation-job-type is set. However, it does add extra confidence - just thinking about how long our test suite already takes. Up to you

HeleneW-dot (changes requested):
I have a whole bunch of questions, most of them only minor/nits and questions for my understanding. 
Requesting changes for one particular question regarding whether our checks on the existence of the  `legacyId` variable is enough, or whether we also need to check if its empty. All other comments are more minor/up to you.

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
tasso94: > any variable referenced in the feel expression must exist

No. If `legacyId` doesn't exist, the expression `=if legacyId then "migrator" else "noop"` resolves to `noop`. 
Let me add another test case validating this.

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
tasso94: I don't get this question. Are you proposing using the C8 ID instead of the C8 key?

Keys in C8 are automatically generated like IDs are in C7.

qa/src/test/java/io/camunda/migrator/qa/CustomJobTypeValidationTest.java
tasso94: Will try and am happy if this makes the assertions simpler.

tasso94: How can `logs.assertContains()` work with an expression like `\\[(\\d+)\\]\\`?

README.md
```diff
@@ -182,16 +182,85 @@ camunda.client:
...
+  validation-job-type: '=if legacyId then "migrator" else "noop"'  # Used for validation with FEEL expression
+```
+
+**Important Notes:**
+- When `validation-job-type` is not defined, `job-type` is used for both purposes
+- The `job-type` is always used for job activation
+- The `validation-job-type` can be a FEEL expression (starts with `=`)
+- Set `validation-job-type` to `DISABLED` to disable job type validation entirely
+- Use FEEL expressions only for validation, not for job activation since on job activation, the FEEL expression is already evaluated to a static value
+- Variables referenced in FEEL expressions must exist in the process instance context
```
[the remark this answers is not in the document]
HeleneW-dot: Yes, my question was more regarding this Feel constraint that says variables referenced in the expression must exist in the instance. If we are recommending an expression like `=if legacyId then "migrator" else "noop"` then that references `legacyId`, so the external instance should have the variable too, but it would be empty I am assuming.

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
HeleneW-dot: Ah right. Then I misunderstood the constraint we mention in the docs:
`Variables referenced in FEEL expressions must exist in the process instance context` Can you clarify?

README.md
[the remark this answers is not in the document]
tasso94: I see your point. The docs are misleading here. I'll rephrase it. The variable must not exist.

core/src/main/java/io/camunda/migrator/RuntimeMigrator.java
HeleneW-dot: > Are you proposing using the C8 ID instead of the C8 key?
No this already ansered my question :) I was more wondering about terminology, since we already learned the definition key / ID equivalency in C7/C8, just hadnt realised the same is true for instance terminology

qa/src/test/java/io/camunda/migrator/qa/CustomJobTypeValidationTest.java
HeleneW-dot: Ah I see, if we have to use that type of matching then probably your way is better.

PR #123, opened 2025-07-15
fix(test): Make WithSpringProfile extendable and other fixes

joaquinfelici:
Related to https://github.com/camunda/camunda-bpm-platform/issues/5262

Allow a class annotated with `@WithSpringProfile` to extend a class also annotated with `@WithSpringProfile` by merging the two profiles.


tasso94 (approved):
Cool stuff!

PR #129, opened 2025-07-21
fix(mapping): fix id/key mapping for process and decision definitions

HeleneW-dot:


tasso94 (approved):
Nice catches! 👍

HeleneW-dot:
rebasing on fixed main before merge

PR #144, opened 2025-07-23
chore(core): adapt latest RDBMS changes from snapshot

tasso94:


core/src/main/java/io/camunda/migrator/config/mybatis/C8Configuration.java
```diff
@@ -379,7 +372,7 @@ public RdbmsWriterFactory rdbmsWriterFactory(
         purgeMapper,
         userTaskMapper,
         variableMapper,
-        rdbmsWriterMetrics,
+        null,
```
tasso94, line 375: Not sure what's wrong here. Removing it for now.

PR #148, opened 2025-07-25
fix(core): should skip multiInstance serviceTasks

venetrius:
related to: camunda/camunda-bpm-platform/issues/5300

### DIFF
--- a/core/src/main/java/io/camunda/migrator/impl/RuntimeValidator.java
+++ b/core/src/main/java/io/camunda/migrator/impl/RuntimeValidator.java
@@ -86,10 +86,15 @@ public class RuntimeValidator {
    */
   public void validateC7FlowNodes(String processDefinitionId, String activityId) {
     BpmnModelInstance c7BpmnModelInstance = repositoryService.getBpmnModelInstance(processDefinitionId);
+    String multiInstanceBodySuffix = "#multiInstanceBody";
+
     FlowElement element = c7BpmnModelInstance.getModelElementById(activityId);
-    if ((element instanceof Activity activity)
-        && (activity.getLoopCharacteristics() instanceof MultiInstanceLoopCharacteristics)) {
-      throw new IllegalStateException(String.format(MULTI_INSTANCE_LOOP_CHARACTERISTICS_ERROR, element.getId()));
+    boolean isMultiInstanceBody = activityId.endsWith(multiInstanceBodySuffix);
+    boolean hasMultiInstanceCharacteristics = element instanceof Activity activity
+        && activity.getLoopCharacteristics() instanceof MultiInstanceLoopCharacteristics;
+
+    if (isMultiInstanceBody || hasMultiInstanceCharacteristics) {
+      throw new IllegalStateException(String.format(MULTI_INSTANCE_LOOP_CHARACTERISTICS_ERROR, activityId));
     }
   }
 
--- a/core/src/main/java/io/camunda/migrator/impl/RuntimeValidator.java
+++ b/core/src/main/java/io/camunda/migrator/impl/RuntimeValidator.java
@@ -8,6 +8,7 @@
 package io.camunda.migrator.impl;
 
 import static io.camunda.migrator.constants.MigratorConstants.LEGACY_ID_VAR_NAME;
+import static io.camunda.migrator.impl.util.C7Utils.MULTI_INSTANCE_BODY_SUFFIX;
 import static io.camunda.migrator.impl.util.C7Utils.getActiveActivityIdsById;
 import static io.camunda.migrator.impl.util.ExceptionUtils.callApi;
 import static io.camunda.migrator.impl.logging.RuntimeValidatorLogs.FAILED_TO_PARSE_BPMN_MODEL;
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/camunda/migrator/impl/RuntimeValidator.java
```diff
@@ -84,10 +84,15 @@ public void validateProcessInstanceState(String legacyRootProcessInstanceId, Con
    */
   public void validateC7FlowNodes(String processDefinitionId, String activityId) {
     BpmnModelInstance c7BpmnModelInstance = repositoryService.getBpmnModelInstance(processDefinitionId);
+    String multiInstanceBodySuffix = "#multiInstanceBody";
```
HeleneW-dot, line 87: Nit
We already have a constant for this in `C7Utils.MULTI_INSTANCE_BODY_SUFFIX`

HeleneW-dot (changes requested):
So this looks good to me, requesting changes because we already had similar validation implemented [here](https://github.com/camunda/c7-data-migrator/blob/4b0718f57f607756a181614e9c0ebb7070120b07/core/src/main/java/io/camunda/migrator/impl/util/C7Utils.java#L30) and I think it would make sense to align the two approaches with this PR. Probably makes sense to have it all in the validator just for consistency

core/src/main/java/io/camunda/migrator/impl/util/C7Utils.java
```diff
@@ -27,7 +27,7 @@ public static Map<String, FlowNode> getActiveActivityIdsById(ActivityInstance ac
     Arrays.asList(activityInstance.getChildActivityInstances()).forEach(actInst -> {
       activeActivities.putAll(getActiveActivityIdsById(actInst, activeActivities));
 
-      if (!SUB_PROCESS_ACTIVITY_TYPE.equals(actInst.getActivityType()) && !actInst.getActivityId().endsWith(MULTI_INSTANCE_BODY_SUFFIX)) {
```
venetrius, line 30: This is part of `getActiveActivityIdsById`.
`getActiveActivityIdsById` is used in 2 none test functions: 
- `activateMigratorJobs` -> this is called after validation, the PI has to be skipped already
- `validateProcessInstanceState` -> this is called just before validation, we want to retrieve multi instance activities, otherwise we could not validate them

core/src/main/java/io/camunda/migrator/impl/RuntimeValidator.java
venetrius: Updated

core/src/main/java/io/camunda/migrator/impl/util/C7Utils.java
HeleneW-dot: Do we also want to move the sub process check in line 30 into the validator? If so, we could open a follow up for that

venetrius: A subprocess is supported by migration there is some difference in how we activate it. This logic should stay here.

core/src/main/java/io/camunda/migrator/impl/RuntimeValidator.java
```diff
@@ -87,9 +88,9 @@ public void validateProcessInstanceState(String legacyRootProcessInstanceId, Con
   public void validateC7FlowNodes(String processDefinitionId, String activityId) {
     BpmnModelInstance c7BpmnModelInstance = repositoryService.getBpmnModelInstance(processDefinitionId);
     FlowElement element = c7BpmnModelInstance.getModelElementById(activityId);
-    if ((element instanceof Activity activity)
-        && (activity.getLoopCharacteristics() instanceof MultiInstanceLoopCharacteristics)) {
-      throw new IllegalStateException(String.format(MULTI_INSTANCE_LOOP_CHARACTERISTICS_ERROR, element.getId()));
+
+    if (isMultiInstanceActivity(activityId, element)) {
+      throw new IllegalStateException(String.format(MULTI_INSTANCE_LOOP_CHARACTERISTICS_ERROR, activityId));
```
tasso94, line 93: The log statement is still not ideal:
```
Skipping process instance with legacyId [c37b0d8d-63e4-11f0-8c26-5682f3a431b7]: Found multi-instance loop characteristics for flow node with id [ServiceTask_1#multiInstanceBody] in C7 process instance.
```

The user searches for the flow node ID `ServiceTask_1#multiInstanceBody` in the BPMN model but cannot find it since the `#multiInstanceBody` is virtual and doesn't actually exist in the BPMN model. I would vote for removing the `#multiInstanceBody` from the activity ID in the log statement.

tasso94 (commented):
One more remark.

core/src/main/java/io/camunda/migrator/impl/RuntimeValidator.java
```diff
@@ -186,6 +187,19 @@ private io.camunda.zeebe.model.bpmn.BpmnModelInstance parseBpmnModel(String xmlS
...
+   */
+  private boolean isMultiInstanceActivity(String activityId, FlowElement element) {
+    boolean isMultiInstanceBody = activityId.endsWith(MULTI_INSTANCE_BODY_SUFFIX);
+    if (isMultiInstanceBody) {
+      return true;
+    }
+    boolean hasMultiInstanceCharacteristics = element instanceof Activity activity
+        && activity.getLoopCharacteristics() instanceof MultiInstanceLoopCharacteristics;
+    return hasMultiInstanceCharacteristics;
+  }
```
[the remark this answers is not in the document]
tasso94: A sequential multi-instance task might actually be supported. The reasoning for skipping parallel is that 5 token detected in C7 on a multi-instance parallel task leads to much more multi-instances in C8. Maybe we can create a follow-up issue to investigate this.

core/src/main/java/io/camunda/migrator/impl/RuntimeValidator.java
venetrius: Created PR: https://github.com/camunda/camunda-7-to-8-data-migrator/pull/180

PR #150, opened 2025-07-25
feat(history): introduce skipping mechanism in history migrator

joaquinfelici:


core/src/main/resources/mapper/IdKey.xml
```diff
@@ -9,19 +9,18 @@
 <!DOCTYPE mapper PUBLIC "-//mybatis.org//DTD Mapper 3.0//EN" "http://mybatis.org/dtd/mybatis-3-mapper.dtd">
 <mapper namespace="io.camunda.migrator.impl.persistence.IdKeyMapper">
 
-  <select id="findSkippedCount" resultType="long">
-    SELECT COUNT(ID) FROM (
-      <include refid="findSkippedByQueryCriteriaSql"/>
-    )
+  <select id="countSkippedByType" resultType="long">
+    SELECT COUNT(ID) FROM ${prefix}MIGRATION_MAPPING
```
joaquinfelici, line 13: Directly using COUNT FROM instead of doing COUNT FROM ( SELECT ... FROM ) should be more performant.

core/src/main/java/io/camunda/migrator/impl/clients/DbClient.java
```diff
@@ -47,6 +50,13 @@ public boolean checkExists(String legacyProcessInstanceId) {
     return callApi(() -> idKeyMapper.checkExists(legacyProcessInstanceId), FAILED_TO_CHECK_EXISTENCE + legacyProcessInstanceId);
   }
 
+  /**
+   * Checks if a process instance exists in the mapping table.
```
joaquinfelici, line 54:
```suggestion
   * Checks if an entity has a non null key in the mapping table
```

core/src/main/java/io/camunda/migrator/HistoryMigrator.java
```diff
@@ -179,258 +192,330 @@ private void migrateDecisionDefinitions() {
...
-            dbClient.insert(legacyIncidentId, legacyIncident.getCreateTime(), dbModel.incidentKey(),
-                HISTORY_INCIDENT);
-            LOGGER.debug("Migration of legacy incident with id [{}] completed.", legacyIncidentId);
-          } else {
-            LOGGER.debug("Migration of legacy incident with id [{}] skipped. Process instance not yet available.",
-                legacyIncidentId);
-          }
+  private void migrateHistoricIncident(HistoricIncident legacyIncident) {
+    String legacyIncidentId = legacyIncident.getId();
+    if (shouldMigrate(legacyIncidentId)) {
```
venetrius, line 322: 🤔 We could early return if entity is already migrated / skipped (and not retry more)

core/src/main/java/io/camunda/migrator/HistoryMigrator.java
```diff
@@ -179,258 +192,330 @@ private void migrateDecisionDefinitions() {
...
+        return;
+      }
+
+      if (isMigrated(legacyProcessInstanceId)) {
+        ProcessInstanceEntity processInstance = findProcessInstanceByLegacyId(legacyProcessInstanceId);
+        if (isMigrated(legacyVariable.getActivityInstanceId())) {
+          Long processInstanceKey = processInstance.processInstanceKey();
+          Long scopeKey = findFlowNodeKey(
+              legacyVariable.getActivityInstanceId()); // TODO does this cover scope correctly?
+          if (scopeKey != null) {
```
venetrius, line 395: I wonder if we have to check the whole hierarchy (PD, PI. activity / task, scope) .
If the PD is not migrated then neither will be thePI or the activity / tasks or scope.
Maybe it is enough to only check the scope (and maybe activity / task

venetrius:
I am trying to the a minimum acceptable PR to merge changes and avoid long lived feature branch & merge conflicts.

TODO for same issue but new PR if acceptable: 
- HistoryMigrationSkippingTest
  - Set Log level for whole test class instead of on test basis
  - Test is disabled: shouldNotMigrateIncidentsWhenJobIsSkipped . Synced with team about state of migrating historic jobs.
  - Missing tests for PD and PI
  - Add decision tables related tests after migrateDecisionDefinitions done in [feat(history): migrate historic decision definitions #158](https://github.com/camunda/camunda-7-to-8-data-migrator/pull/158)
- Handle retry mode
- When checking if entity is already migrated / skipped we need to check by `type` & `id` instead of just `id`.

qa/src/test/java/io/camunda/migrator/qa/history/HistoryMigrationSkippingTest.java
```diff
@@ -0,0 +1,383 @@
...
+        ch.qos.logback.classic.Logger logger = (ch.qos.logback.classic.Logger) LoggerFactory.getLogger(HistoryMigrator.class);
+        logger.setLevel(Level.DEBUG);
+        // given state in c7
+        deployer.deployCamunda7Process("userTaskProcess.bpmn");
+        var processInstance = runtimeService.startProcessInstanceByKey("userTaskProcessId");
+        var task = taskService.createTaskQuery().singleResult();
+        taskService.complete(task.getId());
+
+        // and the process instance is manually set as skipped
+        dbClient.insert(processInstance.getId(), null, IdKeyMapper.TYPE.HISTORY_PROCESS_INSTANCE);
```
tasso94, line 94: It's not ideal to break the black box test principle here by calling an internal API. However, I don't have a better idea how to test this more real-world-y.

core/src/main/java/io/camunda/migrator/HistoryMigrator.java
```diff
@@ -500,4 +585,36 @@ private Long findFlowNodeKey(String activityInstanceId) {
...
+
+  protected void saveRecord(String entityId, Long entityKey, IdKeyMapper.TYPE type) {
+    if (RETRY_SKIPPED.equals(mode)) {
+      dbClient.updateKeyById(entityId, entityKey, type);
+    } else if (MIGRATE.equals(mode)) {
+      dbClient.insert(entityId, entityKey, type);
+    }
+  }
+
+    protected void saveRecord(String entityId, Date date  ,Long entityKey, IdKeyMapper.TYPE type) {
```
tasso94, line 608:
```suggestion
    protected void saveRecord(String entityId, Date date, Long entityKey, IdKeyMapper.TYPE type) {
```

core/src/main/java/io/camunda/migrator/HistoryMigrator.java
```diff
@@ -179,258 +192,330 @@ private void migrateDecisionDefinitions() {
...
+        ProcessInstanceEntity processInstance = findProcessInstanceByLegacyId(legacyProcessInstanceId);
+        if (isMigrated(legacyVariable.getActivityInstanceId())) {
+          Long processInstanceKey = processInstance.processInstanceKey();
+          Long scopeKey = findFlowNodeKey(
+              legacyVariable.getActivityInstanceId()); // TODO does this cover scope correctly?
+          if (scopeKey != null) {
+            VariableDbModel dbModel = variableConverter.apply(legacyVariable, processInstanceKey, scopeKey);
+            variableMapper.insert(dbModel);
+            saveRecord(legacyVariableId, legacyVariable.getCreateTime(), dbModel.variableKey(),
+                HISTORY_VARIABLE);
```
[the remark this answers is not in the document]
tasso94: Yes. An entity should be skipped if its parent cannot be migrated yet, and the parent should be checked again via RETRY. Otherwise, we have rows without foreign key relations, and fixing this is not trivial from a design perspective, either.

tasso94 (approved):
Looks good to me as a first iteration.

* When I run our demo data set, I see many questionable logs that might indicate some flaw, but I haven't investigated this more closely.
* Can you please create follow-up tasks for your remaining open points? We can then prioritise them accordingly.
* Can you please create in https://github.com/camunda/camunda-bpm-platform/issues/5261 sub-issues for: 1) Skip, 2) Retry, 3) Print?
  * Can you please link this PR to the 1) skip sub-issue instead of the parent issue #5261?
  * All follow-up tasks should be sub-issues of #5261 or the child issues.

venetrius:
> Looks good to me as a first iteration.
> 
> * When I run our demo data set, I see many questionable logs that might indicate some flaw, but I haven't investigated this more closely.
> * Can you please create follow-up tasks for your remaining open points? We can then prioritise them accordingly.
> * Can you please create in [In Data Migrator History, skip, retry, and list is aligned with runtime camunda-bpm-platform#5261](https://github.com/camunda/camunda-bpm-platform/issues/5261) sub-issues for: 1) Skip, 2) Retry, 3) Print?
>   
>   * Can you please link this PR to the 1) skip sub-issue instead of the parent issue #5261?
>   * All follow-up tasks should be sub-issues of #5261 or the child issues.

Created the follow subtasks and re-linked the this PR to https://github.com/camunda/camunda-bpm-platform/issues/5325

PR #154, opened 2025-07-29
Reflect repo name change in `README`

tasso94:
Context: https://camunda.slack.com/archives/C088ESH34P2/p1753798940509859?thread_ts=1752047330.776359&cid=C088ESH34P2

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![Java Version](https://img.shields.io/badge/Java-21-blue)](https://www.oracle.com/java/technologies/downloads/#java21)
 [![Spring Boot](https://img.shields.io/badge/Spring%20Boot-3.x-brightgreen)](https://spring.io/projects/spring-boot)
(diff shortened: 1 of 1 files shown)


HeleneW-dot (changes requested):
Found two more places that should be adjusted and one optional one:
- git instructions in Option 2: Build from Source
- git instructions in Clone the repository
- Debugging Tips filename (not really important but for consistency)

❓ do we also want to rename the modules? If so we would also need to adjust `Add a dependency on c7-data-migrator-core (scope: provided)` in readme

tasso94:
> ❓ do we also want to rename the modules? If so we would also need to adjust Add a dependency on c7-data-migrator-core (scope: provided) in readme

I created a follow-up task to change the name of the technical artifacts: https://github.com/camunda/camunda-bpm-platform/issues/5321

HeleneW-dot (approved):
Lgtm!

PR #157, opened 2025-07-31
Vulnerability scanning: publish private scratch image on release

tasso94:
related to camunda/camunda-bpm-platform#5159

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -185,3 +185,59 @@ jobs:
             -DlocalCheckout=${{ inputs.IS_DRY_RUN }} \
             -Pcentral-sonatype-publish,central-release \
             -DskipTests=true -Dskip.central.release=${SKIP_REPO_DEPLOY} -Dskip.camunda.release=true
+
+  docker-image:
+    name: Build and Publish Docker Image
+    runs-on: ubuntu-latest
+    needs: release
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -185,3 +185,59 @@ jobs:
...
+          FROM scratch
+          COPY . /
+          EOF
+
+      - name: Build and Push Docker Image
+        uses: docker/build-push-action@v5
+        with:
+          context: ./docker-context
+          push: true
+          tags: registry.camunda.cloud/team-cambpm/c7-to-c8-data-migrator:${{ env.RELEASE_VERSION }}
```
tasso94, line 242: This should be `camunda-7-to-c8-data-migrator`. However, this image is not user-facing and not functional. We use it for our internal vulnerability scanning process. Since I already communicated this image name to the InfoSec team and to avoid some back-and-forth again, I'll just keep it as it is.

pom.xml
```diff
@@ -174,6 +174,13 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-javadoc-plugin</artifactId>
+        <configuration>
```
yanavasileva, line 180: ❓ Shall we put a small comment, why we need it for a future reference?

yanavasileva (approved):
👍 Looks good, didn't fine anything concerning.

yanavasileva (approved):

PR #158, opened 2025-08-01
5307 1 migrate decision definitions

HeleneW-dot:
Depends on https://github.com/camunda/camunda-7-to-8-data-migrator/pull/153

related to https://github.com/camunda/camunda-bpm-platform/issues/5307

Adds migration of historic decision definition data.
Decision instance data will come with a separate PR.


core/src/main/java/io/camunda/migrator/HistoryMigrator.java
```diff
@@ -227,6 +223,59 @@ private void migrateProcessInstance(HistoricProcessInstance legacyProcessInstanc
...
+        }
+      });
+    }
+  }
+
+  private void migrateDecisionDefinitions() {
+    HistoryMigratorLogs.migratingDecisionDefinitions();
+
+    if (RETRY_SKIPPED.equals(mode)) {
+      // TODO: handle retry
```
tasso94, line 249: Now that retry is on main, this needs to be implemented as well.

qa/src/test/java/io/camunda/migrator/qa/history/HistoryMigrationSkippingTest.java
```diff
@@ -14,6 +14,7 @@
 import io.camunda.migrator.impl.clients.DbClient;
 import io.camunda.migrator.impl.persistence.IdKeyMapper;
 import io.camunda.search.entities.ProcessInstanceEntity;
+import io.camunda.search.query.DecisionRequirementsQuery;
```
HeleneW-dot, line 17: It seems most of the changes in these file was due to incorrect formatting (extra indentation being removed by auto format). The new test to review is `shouldSkipDecisionDefinitionWhenDecisionRequirementsIsSkipped`

yanavasileva (approved):
👍 Pre-approve since I have only cosmetic suggestions.

qa/src/test/java/io/camunda/migrator/qa/history/HistoryMigrationSkippingTest.java
yanavasileva: ❓ Could you format the whole class as there are 4 spaces instead of 2 in other tests cases too.

yanavasileva:
Update PR before merge.

PR #164, opened 2025-08-04
fix(core): refactor DB queries to select / update by id & type

venetrius:
related to: camunda/camunda-bpm-platform/issues/5325

#### Reason of change:
- In C7 ids are unique per entity but ids can collide in the migration mapping table which stores records for multiple entity types. See [comment on issue](https://github.com/camunda/camunda-bpm-platform/issues/5325#issuecomment-3160972390).
#### Key Changes
- Database schema update: Modified primary key from single ID column to composite (ID, TYPE) primary key
- Query method refactoring: Updated all database operations to include type parameter alongside ID
- Test improvements: Added comprehensive tests for ID collision scenarios and type-specific operations

### DIFF
--- a/qa/src/test/java/io/camunda/migrator/qa/history/HistoryMigrationSkippingTest.java
+++ b/qa/src/test/java/io/camunda/migrator/qa/history/HistoryMigrationSkippingTest.java
@@ -397,4 +397,40 @@ public class HistoryMigrationSkippingTest extends HistoryMigrationAbstractTest {
     assertThat(searchHistoricDecisionDefinitions("simpleDmnWithReqs2Id")).isEmpty();
     assertThat(dbClient.countSkippedByType(IdKeyMapper.TYPE.HISTORY_DECISION_DEFINITION)).isEqualTo(2);
   }
+
+    @Test
+    public void shouldNotSkipTaskVariablesWhenEntityWithSameIdButDifferentTypeIsSkipped() {
+        // given state in c7
+        deployer.deployCamunda7Process("userTaskProcess.bpmn");
+        runtimeService.startProcessInstanceByKey("userTaskProcessId");
+
+        var taskId = taskService.createTaskQuery().singleResult().getId();
+
+        // Simulate ID collision by manually inserting a record with the same ID as the task
+        // but with a different type (HISTORY_INCIDENT)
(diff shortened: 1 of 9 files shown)


qa/src/test/java/io/camunda/migrator/qa/history/HistoryMigrationSkippingTest.java
```diff
@@ -397,4 +397,40 @@ public void shouldSkipDecisionDefinitionWhenDecisionRequirementsIsSkipped() {
...
+        var userTasks = searchHistoricUserTasks(processInstanceKey);
+        assertThat(userTasks)
+            .as("User task should be migrated despite ID collision with HISTORY_INCIDENT")
+            .hasSize(1);
+        assertThat(userTasks.getFirst().elementId())
+            .as("User task should have correct id")
+            .isEqualTo("userTaskId");
+
+        // 3. Verify no skip messages in logs for the task
+        logs.assertDoesNotContain("Migration of legacy user task with id [" + taskId + "] skipped");
```
yanavasileva, line 434: ❌ Isn't this wrong assert text, is it? The log is something like: "Migration of historic user task with legacyId [{}] skipped." I suggest to remove the check as it's not reliable, if the log message changes.

qa/src/test/java/io/camunda/migrator/qa/runtime/SkipAndRetryProcessInstancesTest.java
```diff
@@ -49,6 +49,9 @@ class SkipAndRetryProcessInstancesTest extends RuntimeMigrationAbstractTest {
   @Autowired
   private TaskService taskService;
 
+  @Autowired
+  private io.camunda.migrator.HistoryMigrator historyMigrator;
```
yanavasileva, line 53: 🔧 Add import instead:
```suggestion
  private HistoryMigrator historyMigrator;
```

yanavasileva (changes requested):
👍 Looks good. I have comments only on the tests.
✍️ I suggest to add in the PR description or the ticket two sentences what is the use case that this PR resolves. Copilot has good input, I missing the following:
> This PR refactors the database queries to select and update records by both ID and type, addressing ID collision issues where different entity types share the same ID value.
...

`The C7 ids are unique per entity/table but can collide in the migration mapping table so pair ID +type makes them unique.`

PR #171, opened 2025-08-05
feat(core): list skipped historicEntities

venetrius:
related to: camunda/camunda-bpm-platform/issues/5327

### DIFF
--- a/core/src/main/java/io/camunda/migrator/HistoryMigrator.java
+++ b/core/src/main/java/io/camunda/migrator/HistoryMigrator.java
@@ -11,6 +11,7 @@ import static io.camunda.migrator.MigratorMode.LIST_SKIPPED;
 import static io.camunda.migrator.MigratorMode.MIGRATE;
 import static io.camunda.migrator.MigratorMode.RETRY_SKIPPED;
 import static io.camunda.migrator.impl.persistence.IdKeyMapper.TYPE.HISTORY_DECISION_DEFINITION;
+import static io.camunda.migrator.impl.persistence.IdKeyMapper.TYPE.HISTORY_DECISION_INSTANCE;
 import static io.camunda.migrator.impl.persistence.IdKeyMapper.TYPE.HISTORY_FLOW_NODE;
 import static io.camunda.migrator.impl.persistence.IdKeyMapper.TYPE.HISTORY_INCIDENT;
 import static io.camunda.migrator.impl.persistence.IdKeyMapper.TYPE.HISTORY_PROCESS_DEFINITION;
@@ -145,7 +146,8 @@ public class HistoryMigrator {
         HISTORY_USER_TASK,
         HISTORY_VARIABLE,
         HISTORY_INCIDENT,
-        HISTORY_DECISION_DEFINITION
+        HISTORY_DECISION_DEFINITION,
+        HISTORY_DECISION_INSTANCE
     );
 
     historyTypes.forEach(this::printSkippedEntitiesForType);
--- a/core/src/main/java/io/camunda/migrator/HistoryMigrator.java
+++ b/core/src/main/java/io/camunda/migrator/HistoryMigrator.java
@@ -52,9 +52,11 @@ import io.camunda.search.entities.FlowNodeInstanceEntity;
 import io.camunda.search.entities.ProcessDefinitionEntity;
 import io.camunda.search.entities.ProcessInstanceEntity;
 import io.camunda.search.filter.FlowNodeInstanceFilter;
-import java.util.Arrays;
 import java.util.Date;
+import java.util.EnumSet;
 import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
 import org.camunda.bpm.engine.history.HistoricActivityInstance;
 import org.camunda.bpm.engine.history.HistoricIncident;
 import org.camunda.bpm.engine.history.HistoricProcessInstance;
@@ -68,6 +70,11 @@ import org.springframework.stereotype.Component;
 @Component
 @Conditional(C8DataSourceConfigured.class)
 public class HistoryMigrator {
+
+  public static final Set<IdKeyMapper.TYPE> HISTORY_TYPES = EnumSet.allOf(IdKeyMapper.TYPE.class)
+      .stream()
+      .filter(type -> type.name().startsWith("HISTORY"))
+      .collect(Collectors.toCollection(() -> EnumSet.noneOf(IdKeyMapper.TYPE.class)));
 
   // Mappers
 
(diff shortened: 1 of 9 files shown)


core/src/main/java/io/camunda/migrator/HistoryMigrator.java
```diff
@@ -136,6 +138,26 @@ public void start() {
     }
   }
 
+  private void printSkippedHistoryEntities() {
+    List<IdKeyMapper.TYPE> historyTypes = List.of(
```
joaquinfelici, line 142: 🔧  Two things:

1. This can be a constant, no need to create it on every call.
2. We can directly create a list out of an enum like this: `Arrays.asList(IdKeyMapper.TYPE.values())` (in which case we'd need to filter them to only keep the ones that start with `HISTORY`).

core/src/main/java/io/camunda/migrator/impl/util/PrintUtils.java
```diff
@@ -7,20 +7,29 @@
...
 public class PrintUtils {
 
   protected static final Logger PRINTER = LoggerFactory.getLogger("PRINTER");
 
-  public static final String PREVIOUSLY_SKIPPED_INSTANCES_MESSAGE = "Previously skipped process instances:";
-  public static final String NO_SKIPPED_INSTANCES_MESSAGE = "No process instances were skipped during previous migration";
+  public static void printSkippedInstancesHeader(long count, TYPE entityType) {
+    String entityName = entityType.getDisplayName();
+    String message = count > 0
+        ? "Previously skipped " + entityName + "s:"
```
joaquinfelici, line 22: 🔧 Maybe we can put the "entity name" between brackets and make make it capital letters to emphasize that it's an entity name and not just phrasing?

Like instead of :

`Previously skipped historic decision definitions:`

Something like:

`Previously skipped [Historic Decision Definitions]:`

qa/src/test/java/io/camunda/migrator/qa/history/HistoryMigrationListSkippedTest.java
```diff
@@ -0,0 +1,285 @@
...
+    protected DbClient dbClient;
+
+    @Autowired
+    private ManagementService managementService;
+
+    @Autowired
+    protected HistoryService historyService;
+
+    @Test
+    public void shouldListAllSkippedHistoryEntities(CapturedOutput output) {
```
joaquinfelici, line 47:
```suggestion
    public void shouldListAllSkippedHistoricEntities(CapturedOutput output) {
```

qa/src/test/java/io/camunda/migrator/qa/history/HistoryMigrationListSkippedTest.java
```diff
@@ -0,0 +1,285 @@
...
+     * Checks if a line is a log message that should be filtered out in CI environments.
+     * This is necessary because CI may output debug/info logs mixed with actual entity IDs.
+     */
+    private boolean isLogLine(String line) {
+        return line.matches("^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}\\.\\d{3}Z.*") || // ISO timestamp
+               line.matches(".*(DEBUG|INFO|WARN|ERROR).*") || // Log levels
+               line.matches(".*(==>|<==|Preparing:|Parameters:|Total:).*"); // SQL logging patterns
+    }
+
+    private static class EntitySection {
```
joaquinfelici, line 276: ❓ I have the impression that a few methods in here can be extracted and reused by other tests, what do you think?

core/src/main/java/io/camunda/migrator/impl/util/PrintUtils.java
```diff
@@ -7,20 +7,29 @@
...
+        : "No " + entityName + "s were skipped during previous migration";
+    print(message);
+  }
 
+  // Keep the old method for backward compatibility
```
joaquinfelici, line 27: ❓ Do we actually need backwards compatibility since this is still not released?

joaquinfelici (changes requested):
Looks mostly good, added a few comments and questions.

core/src/main/java/io/camunda/migrator/HistoryMigrator.java
venetrius: Updated to use `Arrays.asList(IdKeyMapper.TYPE.values()` and filter out Runtime Process Instances

core/src/main/java/io/camunda/migrator/impl/util/PrintUtils.java
venetrius: Removed

core/src/main/java/io/camunda/migrator/HistoryMigrator.java
```diff
@@ -139,18 +139,9 @@ public void start() {
...
-        HISTORY_USER_TASK,
-        HISTORY_VARIABLE,
-        HISTORY_INCIDENT,
-        HISTORY_DECISION_DEFINITION,
-        HISTORY_DECISION_INSTANCE
-    );
-
-    historyTypes.forEach(this::printSkippedEntitiesForType);
+    Arrays.stream(IdKeyMapper.TYPE.values())
+        .filter(type -> type != IdKeyMapper.TYPE.RUNTIME_PROCESS_INSTANCE)
```
joaquinfelici, line 143: Maybe inclusive filtering is better than exclusive (for example if we add new enums that are not history, we might overlook it and/or need refactoring).

```suggestion
        .filter(type -> type.name().startsWith("HISTORY"))
```

qa/src/test/java/io/camunda/migrator/qa/history/HistoryMigrationListSkippedTest.java
```diff
@@ -126,27 +126,28 @@ private void verifyEntitiesMarkedAsSkipped() {
...
                                            String processDefinitionId, List<String> processInstanceIds) {
-        // Verify all expected entity types are present
-        assertThat(skippedEntitiesByType).containsKeys(
-            "historic process definitions", "historic process instances", "historic flow nodes",
-            "historic user tasks", "historic variables", "historic incidents",
-            "historic decision definitions", "historic decision instances"
-        );
+        // Verify all expected entity types are present - using enum values like the actual implementation
+        String[] expectedEntityTypes = Arrays.stream(IdKeyMapper.TYPE.values())
+            .filter(type -> type != IdKeyMapper.TYPE.RUNTIME_PROCESS_INSTANCE)
```
joaquinfelici, line 131: Same as above. We can make this a Set constant in the code and then reuse it in the tests so we do the logic only once.

joaquinfelici (approved):
Looks good thanks for the fixes. I added a couple more comments but pre-approving anyway 👍

PR #176, opened 2025-08-07
fix(core): add mapper config for IdKeyModel

venetrius:
related to: camunda/camunda-bpm-platform/issues/5339

### DIFF
--- a/core/src/main/resources/mapper/IdKey.xml
+++ b/core/src/main/resources/mapper/IdKey.xml
@@ -9,13 +9,26 @@
 <!DOCTYPE mapper PUBLIC "-//mybatis.org//DTD Mapper 3.0//EN" "http://mybatis.org/dtd/mybatis-3-mapper.dtd">
 <mapper namespace="io.camunda.migrator.impl.persistence.IdKeyMapper">
 
+  <resultMap id="idKeyResultMap" type="io.camunda.migrator.impl.persistence.IdKeyDbModel">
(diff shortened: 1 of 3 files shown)


core/src/main/resources/mapper/IdKey.xml
```diff
@@ -9,15 +9,28 @@
...
-  <select id="findSkippedByType" resultType="io.camunda.migrator.impl.persistence.IdKeyDbModel">
+  <select id="findSkippedByType" resultMap="idKeyResultMap">
     <include refid="findSkippedByQueryCriteriaSql"/>
   </select>
 
+  <select id="findMigratedByType" resultMap="idKeyResultMap">
+    SELECT ID, INSTANCE_KEY, START_DATE, TYPE FROM ${prefix}MIGRATION_MAPPING
+    WHERE TYPE = #{type} AND INSTANCE_KEY IS NOT NULL
+    <include refid="io.camunda.migrator.Commons.pageSql"/>
+  </select>
```
tasso94, line 32: Should we add a comment that we use this SQL mapping in the Cockpit plugin? Just so people know in the future and not just remove because they don't see where it's used?

tasso94 (approved):
Looks good!

core/src/main/resources/mapper/IdKey.xml
venetrius: Good point, I added a comment

PR #199, opened 2025-08-12
feat(core): enable filtering on history list-skipped

venetrius:
related to: camunda/camunda-bpm-platform/issues/5327


core/src/main/java/io/camunda/migrator/impl/persistence/IdKeyMapper.java
```diff
@@ -22,7 +26,7 @@ enum TYPE {
     HISTORY_FLOW_NODE("Historic Flow Node"),
     HISTORY_DECISION_INSTANCE("Historic Decision Instance"),
     HISTORY_DECISION_DEFINITION("Historic Decision Definition"),
-    HISTORY_DECISION_REQUIREMENTS("Historic Decision Requirements"),
+    HISTORY_DECISION_REQUIREMENTS("Historic Decision Requirement"),
```
joaquinfelici, line 29: 1. I guess the enum should change as well to `HISTORY_DECISION_REQUIREMENT` ?
2. Should we add this one to the "Available Entity Types for Filtering" section of the README?

joaquinfelici (approved):
Looks good to me. Added one last comment I missed before, but pre-approving ✅

PR #206, opened 2025-08-22
test(core): test user task mapping in history 

yanavasileva:
https://github.com/camunda/camunda-7-to-8-data-migrator/issues/366


yanavasileva:
> Are historyCleanupDate & partitionId intentionally not tested?
I didn't come up with way to test them. Any suggestions, @venetrius?

venetrius:
> > Are historyCleanupDate & partitionId intentionally not tested?
> 
> I didn't come up with way to test them. Any suggestions, @venetrius?

I see these fields are only present on UserTaskDbModel but not on UserTaskEntity. We can only create a UserTaskDbModel but no way to load it. 
The only option to verify these values would be to ask the rdbms team to expose these fields or to write a utility in qa that directly checks in the rdbms tables.

Let's merge it as it is 👍

PR #207, opened 2025-08-22
feat(core): map historyFlowNodes

venetrius:
related to camunda/camunda-bpm-platform/issues/5005


core/src/main/java/io/camunda/migrator/converter/FlowNodeConverter.java
```diff
@@ -8,57 +8,129 @@
...
+      return FlowNodeInstanceEntity.FlowNodeState.ACTIVE;
+    } else if (legacyFlowNode.isCanceled()) {
+      return FlowNodeInstanceEntity.FlowNodeState.TERMINATED;
+    } else {
+      return FlowNodeInstanceEntity.FlowNodeState.COMPLETED;
+    }
+  }
+
+//  In c7 for example we have event subscriptions
   protected FlowNodeType convertType(String activityType) {
```
venetrius, line 82: convertType is still work in progress, will create a follow up PR in the same issue.

qa/src/test/java/io/camunda/migrator/qa/history/entity/FlowNodeTreePathTest.java
yanavasileva, line 1: 📓 Use case for nested call activities is missing. But we can add it with the camunda-bpm-platform/issues/5359 ticket. Just add a note please.

qa/src/test/java/io/camunda/migrator/qa/history/entity/FlowNodeConverterTest.java
```diff
@@ -0,0 +1,231 @@
...
+      assertThat(flowNodes).hasSizeGreaterThanOrEqualTo(1);
+    } else {
+      assertThat(flowNodes).hasSize(1);
+    }
+
+    // Verify all instances have correct fields (use the first one for detailed verification)
+    FlowNodeInstanceEntity flowNode = flowNodes.getFirst();
+
+    // Verify all converted fields
+    assertThat(flowNode.flowNodeInstanceKey()).isNotNull().isPositive();
```
yanavasileva, line 172: 🔧 Let's do equals check instead of `notNull`.

yanavasileva (changes requested):
👍 Overall looks good.

qa/src/test/java/io/camunda/migrator/qa/history/entity/FlowNodeConverterTest.java
```diff
@@ -0,0 +1,231 @@
...
+    retryAndSucceed(processInstance);
+
+    // Check for incidents
+    var historicIncidents = historyService.createHistoricIncidentQuery().processInstanceId(processInstanceId).list();
+    assertThat(historicIncidents).hasSize(1);
+
+    // when - migrate history
+    historyMigrator.migrate();
+    // writer needs to be flushed for incidents to bewritten into the flow node table
+    rdbmsWriter.flush();
```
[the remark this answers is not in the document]
venetrius: Right now writers are used for: 
1. Updating process instances with treePath info in this test
2. Adding an incident to a flowNode

In real word we also need to flush if we have to make sure the updated info is available on the next read. It is not an issue when creating an entity. For the test it was required as there is no direct writer to update the treePath by id, it is required create a PI model instance with same id and save that. 
For example when adding an incident to a flow node it does not matter when the data is written to the db as we do not rely on reading it later on.

qa/src/test/java/io/camunda/migrator/qa/history/entity/FlowNodeConverterTest.java
```diff
@@ -0,0 +1,231 @@
...
+      }
+    }
+  }
+
+  private List<FlowNodeInstanceEntity> searchHistoricFlowNodes(Long processInstanceKey) {
+    return rdbmsService.getFlowNodeInstanceReader()
+        .search(io.camunda.search.query.FlowNodeInstanceQuery.of(queryBuilder -> queryBuilder.filter(
+            filterBuilder -> filterBuilder.processInstanceKeys(processInstanceKey))))
+        .items();
+  }
```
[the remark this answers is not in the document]
venetrius: removed

qa/src/test/java/io/camunda/migrator/qa/history/entity/FlowNodeTreePathTest.java
venetrius: I did consider that. I think that is not relevant for the flowNodes. The flow node should create the treePath based on their parent scope which is either: 
- Their own PI 
- One of the other flowNodes 

Both of these scenarios are tested. 
In case of a call activity the PI treePath is not just a single key but something like `callingPIKey/CallActivityKey/calledPIKey` but this is not really an edge case.  On the other hand I might be wrong about this one. 
Added a note to https://github.com/camunda/camunda-bpm-platform/issues/5359

yanavasileva:
Docs for limitations: https://github.com/camunda/camunda-7-to-8-data-migrator?tab=readme-ov-file#history-migration

core/src/main/java/io/camunda/migrator/HistoryMigrator.java
```diff
@@ -455,17 +468,64 @@ private void migrateFlowNode(HistoricActivityInstance legacyFlowNode) {
...
+        String parentTreePath = determineParentTreePath(legacyFlowNode, processInstance);
+
+        if (flowNodeScopeKey != null &&  parentTreePath != null) {
+          FlowNodeInstanceDbModel dbModel = flowNodeConverter.apply(legacyFlowNode, processDefinitionKey, processInstanceKey, flowNodeScopeKey, parentTreePath);
+          flowNodeMapper.insert(dbModel);
+          saveRecord(legacyFlowNodeId, legacyFlowNode.getStartTime(), dbModel.flowNodeInstanceKey(), HISTORY_FLOW_NODE);
+          HistoryMigratorLogs.migratingHistoricFlowNodeCompleted(legacyFlowNodeId);
+        } else {
+          saveRecord(legacyFlowNodeId, null, HISTORY_FLOW_NODE);
+          HistoryMigratorLogs.skippingHistoricFlowNodeDueToMissingParent(legacyFlowNodeId);
```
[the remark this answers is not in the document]
venetrius: Updated code to set treePath to null if parent's treePath is null instead of skipping entities.

yanavasileva:
Beside the above, the PR is good to go.

venetrius:
Created a follow up [subtask](https://github.com/camunda/camunda-bpm-platform/issues/5392) to  verifyflow node type conversion

core/src/main/java/io/camunda/migrator/impl/logging/HistoryMigratorLogs.java
```diff
@@ -61,6 +61,7 @@ public class HistoryMigratorLogs {
   public static final String MIGRATING_FLOW_NODE = "Migrating historic flow nodes with legacyId: [{}]";
   public static final String MIGRATING_FLOW_NODE_COMPLETED = "Migration of historic flow nodes with legacyId [{}] completed.";
   public static final String SKIPPING_FLOW_NODE = "Migration of historic flow nodes with legacyId [{}] skipped. Process instance yet not available.";
+  public static final String SKIPPING_FLOW_NODE_MISSING_PARENT = "Migration of historic flow node with legacyId [{}] skipped. Parent flow node not yet available.";
```
yanavasileva, line 64: 🔧 Not used anymore:
```suggestion
```

core/src/main/java/io/camunda/migrator/HistoryMigrator.java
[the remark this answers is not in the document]
tasso94: The fact that this is ordered currently is a temporary situation. The vision foresees that migration should be possible for each entity in parallel. If not skipping, we persist a flow node with `scopeKey` set to `null` and can never fix it.

core/src/main/java/io/camunda/migrator/HistoryMigrator.java
```diff
@@ -415,15 +419,24 @@ private void migrateIncident(HistoricIncident legacyIncident) {
...
         if (processInstanceKey != null) {
           Long flowNodeInstanceKey = findFlowNodeInstanceKey(legacyIncident.getActivityId(), legacyIncident.getProcessInstanceId());
           Long processDefinitionKey = findProcessDefinitionKey(legacyIncident.getProcessDefinitionId());
           Long jobDefinitionKey = null; // TODO Job table doesn't exist yet.
           IncidentDbModel dbModel = incidentConverter.apply(legacyIncident, processDefinitionKey, processInstanceKey, jobDefinitionKey, flowNodeInstanceKey);
           incidentMapper.insert(dbModel);
+          // Register incident with flow node if flow node exists
+          if (flowNodeInstanceKey != null) {
+            FlowNodeInstanceDbModel flowNodeInstance = findFlowNodeByKey(flowNodeInstanceKey);
+            if (flowNodeInstance != null) {
```
tasso94, line 434: Why don't we skip migrating the incident if the flow node instance hasn't yet been migrated?

core/src/main/java/io/camunda/migrator/HistoryMigrator.java
```diff
@@ -559,6 +575,39 @@ private void migrateFlowNode(HistoricActivityInstance legacyFlowNode) {
     }
   }
 
+  private String determineParentTreePath(HistoricActivityInstance legacyFlowNode, ProcessInstanceEntity processInstance) {
+    String parentActivityInstanceId = legacyFlowNode.getParentActivityInstanceId();
+    if (parentActivityInstanceId == null || parentActivityInstanceId.equals(legacyFlowNode.getProcessInstanceId())) {
+      return processInstance.treePath();
```
tasso94, line 581: Do we construct a `treePath` for process instances currently?

PR #228, opened 2025-09-10
fix(core): add variables check to c8 modification

yanavasileva:
https://github.com/camunda/camunda-bpm-platform/issues/5396


yanavasileva:
I couldn't come up with a JUnit test, I test it manually with alpha8.

PR #229, opened 2025-09-10
revert(core): revert incident data for HPI

yanavasileva:
https://github.com/camunda/camunda-bpm-platform/issues/5400


venetrius (approved):
👍 Revert looks good

PR #236, opened 2025-09-17
feat(core): add --force flag to drop migration schema in all cases

joaquinfelici:


core/src/main/java/io/camunda/migrator/config/mybatis/SchemaShutdownCleaner.java
```diff
@@ -45,15 +49,23 @@ public class SchemaShutdownCleaner {
   @Autowired
   protected DbClient dbClient;
 
+  @Autowired
+  protected Environment environment;
+
   @PreDestroy
   public void cleanUp() {
```
yanavasileva, line 56: 🙃 For logs we use `XXXLogs` classes in most places, so you can move them out from this class, up to you.

qa/src/test/java/io/camunda/migrator/qa/persistence/DropSchemaTest.java
```diff
@@ -55,45 +56,68 @@ void shouldMigrationSchemaBeKeptOnShutdown() throws Exception {
   @Test
   void shouldMigrationSchemaBeDroppedOnShutdown() throws Exception {
```
yanavasileva, line 57: 🔧 include prefix in the name
```suggestion
  void shouldMigrationSchemaWithPrefixBeDroppedOnShutdown() throws Exception {
```

qa/src/test/java/io/camunda/migrator/qa/persistence/DropSchemaTest.java
```diff
@@ -55,45 +56,68 @@ void shouldMigrationSchemaBeKeptOnShutdown() throws Exception {
...
     // when application is shut down
     context.close();
 
-    // then migration schema is dropped
-    assertThat(tableExists(durableDataSource, "FOO_" + MIGRATION_MAPPING_TABLE)).isFalse();
+    // then migration schema is kept
+    assertThat(tableExists(durableDataSource, MIGRATION_MAPPING_TABLE)).isTrue();
+
+    // cleanup db after text
+    clearMigrationMappingTable(durableDataSource, "");
```
yanavasileva, line 88: 🔧 If the test fails, this cleanup won't be executed. Consider moving it to `@AfterEach`.

qa/src/test/java/io/camunda/migrator/qa/persistence/DropSchemaTest.java
yanavasileva, line 1: ✍️ Test case for only `force` flag passed without `drop-schema` is missing.

yanavasileva (commented):
👍 Implementation looks good to me. I have suggestion for one more test case.

qa/src/test/java/io/camunda/migrator/qa/persistence/DropSchemaTest.java
joaquinfelici: Done

yanavasileva (commented):
👍 Nice. Thank you for considering my feedback.

PR #263, opened 2025-09-23
chore(distro): make flags mutual exclusive

yanavasileva:
https://github.com/camunda/camunda-bpm-platform/issues/5417

### DIFF
--- a/qa/src/test/java/io/camunda/migrator/qa/distribution/DistributionSmokeTest.java
+++ b/qa/src/test/java/io/camunda/migrator/qa/distribution/DistributionSmokeTest.java
@@ -61,7 +61,7 @@ class DistributionSmokeTest {
   }
 
   @Test
-  @Timeout(value = 60, unit = TimeUnit.SECONDS)
+  @Timeout(value = 30, unit = TimeUnit.SECONDS)
   void shouldShowUsageWhenInvalidFlagProvided() throws Exception {
     // given
     ProcessBuilder processBuilder = createProcessBuilder("--runtime", "--invalid-flag");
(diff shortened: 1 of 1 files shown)


distro/src/main/java/io/camunda/migrator/app/MigratorApp.java
```diff
@@ -104,23 +105,30 @@ protected static void validateArguments(String[] args) {
...
 
+    // Check if we have too many flags (not counting entity type parameters)
+    if (flagCount > MAX_FLAGS) {
+      throw new IllegalArgumentException("Error: Too many arguments.");
+    }
+
+    // Validate skipped migration configuration
+    if (hasListSkipped && hasRetrySkipped) {
+      throw new IllegalArgumentException(
+          "Conflicting flags: --list-skipped and --retry-skipped cannot be used together. Please specify only one migration mode.");
```
HeleneW-dot, line 116: ❓ Question just for clarity on terminology: for me I assumed migration mode means history or runtime and the other flags are just adjustments of the behaviour for each mode - do we have any agreed terminlogy here or is any flag also a migrator mode?

HeleneW-dot, line 116: Nit: we could move this to above the for each loop to avoid having to do the loop if the combination is already invalid

yanavasileva: I will change it to avoid confusion

qa/src/test/java/io/camunda/migrator/qa/distribution/DistributionSmokeTest.java
```diff
@@ -112,7 +112,7 @@ void shouldShowUsageWhenHelpFlagProvided() throws Exception {
   @Timeout(value = 60, unit = TimeUnit.SECONDS)
   void shouldShowUsageWhenTooManyArgumentsProvided() throws Exception {
     // given
-    ProcessBuilder processBuilder = createProcessBuilder("--runtime", "--history", "--history", "--drop-schema", "--force", "--list-skipped", "--retry-skipped");
+    ProcessBuilder processBuilder = createProcessBuilder("--runtime", "--history", "--history", "--drop-schema", "--force", "--list-skipped");
```
yanavasileva, line 115: With the new order of validation list and retry cause another failure first.

PR #268, opened 2025-09-23
fix(plugins/cockpit): revert renovate node update

venetrius:
related to: camunda/camunda-bpm-platform/issues/5430


venetrius:
Reverted changes from Renovate and did a smoke tests.

PR #277, opened 2025-09-25
chore(test): use CPT to assert on local variables

joaquinfelici:
Related to https://github.com/camunda/camunda-bpm-platform/issues/5251


yanavasileva (approved):
👍

PR #299, opened 2025-10-06
fix(core): skip PI if active gateway

venetrius:
related to: camunda/camunda-bpm-platform/issues/5455

## Related issues

closes #

### DIFF
--- a/core/src/main/java/io/camunda/migrator/impl/RuntimeValidator.java
+++ b/core/src/main/java/io/camunda/migrator/impl/RuntimeValidator.java
@@ -66,8 +66,7 @@ public class RuntimeValidator {
   /**
    * Validates C7 flow nodes for multi-instance loop characteristics.
    */
-  public void validateC7FlowNodes(String processDefinitionId, String processInstanceId, FlowNode activity) {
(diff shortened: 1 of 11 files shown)


tasso94 (approved):
Looks good to me. 

This breaks support for async parallel splitting gateways if the job has not yet been executed. However, this situation might be unlikely.

Another workaround could have been to resolve the deadlocked process instance later in Operate using modification. With this solution, it is no longer possible, since PIs are now marked as skipped. If we receive feedback on this, let's consider:
1. Allowing force migration, which disables validation for specific or all process instances.
2. Disable certain validation rules via configuration

venetrius:
> This breaks support for async parallel splitting gateways if the job has not yet been executed. However, this situation might be unlikely.

It is really unlikely that a process instance is stopped at the time that the token is at an async splitting gateway, but could happen. This change would skip that PI as well. 
I think with the current state of skipping and retrying such incident is easy to catch and remedy by the user ( if extremely rare). 

 Updated the [follow up ticket](https://github.com/camunda/camunda-bpm-platform/issues/5461) with the idea to let the migration continue with deadlock.

PR #369, opened 2025-10-30
chore(github): adjust issue templates

tasso94:
## Description

Adjust issue templates.

## Related issues

related to camunda/team-automation-platform#176


yanavasileva (approved):
Looks good. Two improvement ideas:
1. The extension of the filea can be changed to `.md` only or `.yml` so the Preview option in Github is available for the files. example: [link](https://github.com/camunda/camunda-bpm-platform/tree/master/.github/ISSUE_TEMPLATE)
2. If we want to have the types of tickets in a specific order, we can put numbers as prefix of the file names. Otherwise the are ordered alphabetically. ref: [docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#changing-the-order-of-templates)

PR #447, opened 2025-11-25
414 merge code conversion tooling

venetrius:
## Description
related to: camunda/camunda-7-to-8-migration-tooling/issues/414

## Related issues

closes #


venetrius:
Re created in https://github.com/camunda/camunda-7-to-8-migration-tooling/pull/449

PR #1461, opened 2026-06-23
feat: migrate C7 business key to C8 business ID during runtime migration [backport 0.3]

### LINKED ISSUE #1359: Business Key Runtime Migration
### User Story (Required on creation)

* As a user migrating from Camunda 7 to Camunda 8, I want Camunda 7 business keys to be migrated to Camunda 8 business IDs so that business-level correlations and identifiers remain available after migration.

### Functional Requirements (Required before implementation)
* Extend the runtime data migrator to support migrating Camunda 7 business keys to Camunda 8 business IDs.
* Preserve the original business key value during migration without requiring manual remapping.
* Apply this only to runtime data migration, since history data migration already covers this case.
* Document the behavior clearly so users understand that business key migration is supported for runtime data.

### Technical Requirements (Required before implementation)
* Identify where the runtime data migrator currently omits business key handling.
* Map the Camunda 7 business key field to the corresponding Camunda 8 business ID field during runtime migration.
* Add automated tests covering runtime migration with and without business keys.
* Verify that the migration remains backward compatible for instances where no business key is set.

### Limitations of Scope
* This request covers runtime data migration only.
* This request does not change or extend history data migration.
* This request does not include any broader migration logic beyond business key to business ID mapping.
    
### Pull Requests

https://github.com/camunda/camunda-7-to-8-migration-tooling/pull/1360

### Dev2QA handover
- [ ] Does this ticket need a QA test, and the testing goals are not clear from the description? Add a [Dev2QA handover comment](https://confluence.camunda.com/display/AP/Handover+Dev+-%3E+Testing)

bojtospeter:
Backport of #1360 to `maintenance/0.3`.

## Changes

- `C8Client.createProcessInstance()` now accepts optional `businessId` param; conditionally calls `.businessId()` only when non-null
- `RuntimeMigrator.startNewProcessInstance()` reads `processInstance.getBusinessKey()` and forwards it as C8 business ID
- Integration test `BusinessKeyMigrationTest` covers null, empty, and present business key cases

### DIFF
--- a/data-migrator/core/src/main/java/io/camunda/migration/data/RuntimeMigrator.java
+++ b/data-migrator/core/src/main/java/io/camunda/migration/data/RuntimeMigrator.java
@@ -178,7 +178,8 @@ public class RuntimeMigrator {
       // Ensure all variables are fetched and can be transformed before starting the new instance
       Map<String, Object> globalVariables = variableService.getGlobalVariables(c7ProcessInstanceId);
 
-      return c8Client.createProcessInstance(bpmnProcessId, processInstance.getTenantId(), globalVariables)
+      return c8Client.createProcessInstance(bpmnProcessId, processInstance.getTenantId(), globalVariables,
+          processInstance.getBusinessKey())
           .getProcessInstanceKey();
     } else {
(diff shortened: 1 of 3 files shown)


venetrius (commented):
👍

PR #2402, opened 2026-08-28
feat(diagram-converter): migrate simple form expressions

### LINKED ISSUE #2368: Diagram converter: form migration parity with BPMN (auto-transform JUEL to FEEL and more)
### User Story (Required on creation)

As a user migrating a Camunda 7 process application to Camunda 8, I want the Diagram Converter to transform `.form` file content automatically where it safely can — the same way it does for BPMN/DMN models — instead of only reporting findings, so that simple cases don't require manual work.

### Background

#1201 added metadata-only `.form` conversion, #2340 added analysis findings for `.form` files. Both deliberately take a conservative approach: findings only, no content transformation.

BPMN conversion, by contrast, does actual transformation work: JUEL expressions are converted to FEEL (`expression-transformation-result` / REVIEW messages with `old -> new`), attributes are mapped, unsupported elements are rewritten where possible. This issue tracks bringing forms to parity: what we do with BPMN, do with forms.

### Functional Requirements (Required before implementation)

* Auto-transform trivially safe JUEL expressions in form component properties (labels, text, default values, ...) to FEEL, e.g.:
  * `${variableName}` -> `= variableName` (simple variable reference)
  * Report the transformation as a REVIEW finding ("Please review transformed expression"), mirroring the BPMN `expression` message pattern
* Keep reporting REVIEW findings for expressions that cannot be transformed safely (method invocations, `execution` context, complex EL), instead of silently transforming them
* Extend the `FormChecker`/conversion pipeline so that conversion and checking share one traversal (today `FormConverter.convert` and `FormChecker.check` walk the JSON separately)
* Evaluate further parity candidates, e.g.:
  * schema version upgrade on conversion (bump `schemaVersion` to the target's expected version)
  * renaming/mapping deprecated component properties where form-js documents a migration path

### Technical Requirements (Required before implementation)

* Decide the JUEL->FEEL transformation approach: reuse the BPMN expression transformation logic where applicable (it already handles simple variable references) or a form-specific subset
* Transformations must be conservative: only provably safe patterns are rewritten; everything else keeps the REVIEW finding
* Regression coverage for transformed and non-transformed expression categories, CLI and webapp

### Limitations of Scope

* Embedded HTML forms and `camunda:formData` generated task forms stay out of scope (not `.form` JSON)
* Complex JUEL (method calls, bean references, `execution`) is not auto-transformable - findings remain the answer there

### Links

* #1201 - metadata-only `.form` conversion
* #2340 - analysis findings for `.form` files (PR #2349)

bojtospeter:
## Summary

- transform exact simple JUEL form values such as `${customerName}` and `#{customerName}` to FEEL while preserving complex expressions for manual review
- collect conversion findings during the same JSON traversal used to mutate form content
- use the combined result in the CLI and webapp, with category-scoped regression coverage across all three layers

## Changes

- add a conservative form processor and combined conversion/check result in `diagram-converter/core`
- reuse the BPMN REVIEW message pattern for successful expression transformations
- preserve schema versions and deprecated properties because no schema-aware, semantics-preserving migration is available
- update CLI and webapp form processing and documentation

## Test plan

- [x] `JAVA_HOME=/opt/homebrew/opt/openjdk@21 mvn verify -f diagram-converter/pom.xml -Pdistro,checkFormat --batch-mode`
- [x] safe `${...}` and `#{...}` references transform in nested form component properties
- [x] method calls, execution context, interpolation, paths, operators, reserved names, and structural properties remain unchanged with REVIEW findings
- [x] CLI reports and webapp check/convert endpoints expose the expected behavior

## Baseline

Built from commit: `1c13ad32fe3abd05363d9bd77fa652b5f92cb0d4`

related to #2368

🤖 Generated with GitHub Copilot


bojtospeter:
@copilot review

